PR #1, opened 2022-04-11
Initial repo setup and basic ContentLoader component

rnathuji:


src/lib/utils.ts
```diff
@@ -0,0 +1,29 @@
+const loadMathJax = async (): Promise<boolean> => {
+  const maybeMoodleYUI = window.Y
+
+  if (maybeMoodleYUI === undefined) {
+    // TODO: Load MathJax manually to support things like preview outside of
+    // Moodle
+    return false
```
rnathuji, line 7: I hit this once and I just observed @mjoss17 experience the same: I believe that sometimes we hit this code path even when we're in Moodle and math ends up not rendering. It must be due to some async loading and ensuing race condition. I'll dig into it a bit more.

src/components/ContentLoader.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const ContentLoader = ({ contentId, contentLoadedCallback }: { contentId: string, contentLoadedCallback: () => void }): JSX.Element => {
+  const [innerHTMLContent, setInnerHTMLContent] = useState<string>('')
+  const [fetchStatus, setFetchStatus] = useState<Number>(FetchStatus.Unfetched)
+
+  const fetchContent = async (): Promise<void> => {
+    const request = new Request(`${ENV.OS_RAISE_CONTENT_URL_PREFIX}/${contentId}.json`)
+
+    try {
+      const response = await fetch(request)
+      const data = await response.json() as ContentResponse
```
mjoss17, line 24: I'm confused about which things are running asynchronously here. Is it the case that multiple of these ContentLoader functional components are making their fetches simultaneously? If that is the case how are these two awaits working together? - do all the components wait on the fetches to finish, then for all the responses to be collected?

src/components/ContentLoader.tsx
```diff
@@ -0,0 +1,55 @@
...
+      setInnerHTMLContent(htmlContent)
+      setFetchStatus(FetchStatus.FetchSuccess)
+    } catch {
+      setFetchStatus(FetchStatus.FetchFailure)
+    }
+  }
+
+  useEffect(() => {
+    fetchContent().catch(() => {})
+  }, [])
```
mjoss17, line 36: Does this `UseEffect` hook enable `fetchContent()` to be called (if so by what), capture its errors, or both? I'm curious very generally about how these hooks are used

src/lib/utils.ts
rnathuji: I wasn't able to reproduce this, but after some code inspection: 1) I believe my initial hunch about this code path being taken was wrong (our script is defer'd and Moodle is loading globals like `M` and `Y` through in-line / synchronously loaded JS); 2) I found where a definite async issue that can occur between MathJax and the `ContentLoader` completing. I just missed this when writing the code, and it happened to randomly work most of the time despite the flawed logic. In any case, [this commit](https://github.com/openstax/k12-apps-raise/pull/1/commits/209a520e165b0cb5c7563dbf1ebcd2e63104c109) should fix things.

src/components/ContentLoader.tsx
rnathuji: It's the first. The usage here is probably best described by the notes [here](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect) on conditionally firing effects. The key thing here is providing `[]` as the dependency input so we know `fetchContent` is only invoked once.

src/lib/content.tsx
```diff
@@ -0,0 +1,40 @@
...
+      return
+    }
+
+    if (htmlElem.hasChildNodes()) {
+      console.log('WARNING: Found non-empty os-raise-content')
+    }
+
+    mathifyElement(elem).catch((error) => {
+      console.error(error)
+    })
```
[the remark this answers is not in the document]
rnathuji: There is some background in [this comment](https://github.com/openstax/k12/issues/64#issuecomment-1073916105), but the general gist is that we are completely outside of Moodle's MathJax filter (e.g. it will do nothing for us due to some of our findings in that issue thread). However, we go ahead and "patch" MathJax into the `window` using Moodle's configuration if we find it (this is what `loadMathJax` is for). The `mathifyElement` is abstracting `MathJax` a bit, but it's implementation of `window.MathJax.Hub.Queue` is generally how you tell MathJax to render something using [Queue()](https://docs.mathjax.org/en/v2.4-latest/api/hub.html#methods).

src/lib/content.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const contentPromise = new Promise<void>((resolve) => {
+      createRoot(htmlElem).render(
+        <React.StrictMode>
+          <ContentLoader contentId={contentId} contentLoadedCallback={resolve}/>
```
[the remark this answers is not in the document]
rnathuji: I'm open to alternate ways of accomplishing this, but the general problem is there are certain things (e.g. like invoking math rendering, parsing for "page types", etc.) we can only do after all of the content has been loaded and reflected in the DOM. Otherwise, for example, when we tell `MathJax` to render math it will find nothing and we'll get unrendered math when the component finally does insert the content in the DOM. This callback is invoked by `ContentLoader` when it is "done" [here](https://github.com/openstax/k12-apps-raise/blob/209a520e165b0cb5c7563dbf1ebcd2e63104c109/src/components/ContentLoader.tsx#L39-L42). The code here uses that invocation to create a Promise that we can wait on a few liens down.

mjoss17 (approved):

PR #2, opened 2022-04-18
Add support for manual MathJax setup and playwright tests

rnathuji:
This PR includes two items:

1. Some code to manually set up MathJax: This is included here as it's necessary for math to render during the playwright tests, but will likely be more generally useful particularly if we end up disabling the MathJax Moodle plugin to manage things on our own.

2. Initial setup of [Playwright](https://playwright.dev/) in the repo along with a couple of test cases.


src/lib/math.ts
```diff
@@ -0,0 +1,83 @@
...
+          oldOnloadCallback()
+        } else {
+          configureMathJax()
+        }
+        resolve()
+      }
+
+      script.onload = chainedCallback
+    })
+  }
```
mjoss17, line 50: I probably need some clarification on what this condition is doing with chained call backs. How does this affect the manner in which mathjax is loaded?

mjoss17 (approved):

PR #4, opened 2022-04-28
Implement ContentBlock and update ContentLoader to blockify

rnathuji:


P-Gill97 (approved):
Tested the ContentBlock.tsx in my moodle local instance. ContentBlock and loader works as expected with an equation. I used a data-wait-for-event and the content did not show 👍.

mjoss17 (approved):

PR #5, opened 2022-04-28
Filter mathifying for only specific delimeters

mjoss17:


rnathuji (commented):
This looks good to me. Let's add the playwright test to this PR and it should be good to go.

src/lib/math.ts
```diff
@@ -70,8 +70,16 @@ export const loadMathJax = async (): Promise<boolean> => {
   }
 }
 
-export const mathifyElement = (elem: Element): void => {
+export const mathifyElement = async (elem: Element): Promise<void> => {
```
rnathuji, line 73: Oh, I think you got the lint error because the merge made this async again. This line should be `export const mathifyElement = (elem: Element): void => {` and you can then back out the change to `ContentBlock.tsx`.

rnathuji (approved):

PR #9, opened 2022-04-29
Add CTA Blocks

mjoss17:


src/components/CTABlock.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const clickHandler = (): void => {
+    if (fireEvent !== undefined) {
+      const clickEvent = new CustomEvent(fireEvent)
+      document.dispatchEvent(clickEvent)
+    }
+    setclicked(true)
+  }
+
+  const promptButton = (): JSX.Element => {
```
rnathuji, line 24: Minor: Can we rename this to something like `maybePromptAndButton`? I otherwise get confused when reading the render block :sweat_smile:.

src/components/EventControlledContent.tsx
```diff
@@ -7,6 +8,15 @@ interface EventControlledContentProps {
 
 export const EventControlledContent = ({ waitForEvent, children }: EventControlledContentProps): JSX.Element => {
   const [shouldRender, setShouldRender] = useState(waitForEvent === undefined)
+  const contentDiv = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
```
rnathuji, line 13: Two things on this:

1. The ref here will always be `null` as it's not actually used anywhere.

2. While we used this approach in the spike, in the "real" iteration let's use a different pattern and not do this in `EventControlledContent` at all. Instead, you can refer to `ContentBlock.tsx` for an alternative pattern we can follow (though open to other suggestions as well)

src/tests/CTABlock.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    </div>
+  )
+  const eventHandler = jest.fn()
+
+  document.addEventListener('Event', eventHandler)
+  fireEvent.click(screen.getByText('Click me!'))
+  expect(eventHandler).toHaveBeenCalled()
+})
+
+test('CTABlock button prompt disappear', async () => {
```
rnathuji, line 32: Minor: For improved readability maybe `CTABlock button and prompt disappear on click`

rnathuji (commented):
A few minor comments in-line, but I think this PR is generally missing the following:

1. An e2e playwright test would be nice here. Perhaps some variant of the ones [here](https://github.com/openstax/k12-apps-raise/blob/interactive-block-spikes/specs/content-templates.spec.ts) (without the tooltips and just focusing on the conditional rendering + math).

2. The test cases :point_up: will fail until the following are added:
    * Add the new block type to [this check](https://github.com/openstax/k12-apps-raise/blob/55c1e5dda6440df7a1f1b760eba7a40453281cd3/src/lib/blocks.tsx#L8)
    * Add the new block type to [blockifyElement](https://github.com/openstax/k12-apps-raise/blob/55c1e5dda6440df7a1f1b760eba7a40453281cd3/src/lib/blocks.tsx#L54) by writing a `parseCTABlock` function

3. For the Moodle storage path we'll want to add a `renderCTABlocks` function [here](https://github.com/openstax/k12-apps-raise/blob/55c1e5dda6440df7a1f1b760eba7a40453281cd3/src/main.tsx#L12)

4. You may want to refer to `ContentBlock.test.tsx` for test cases that might be relevant for `CTABlock.test.tsx`, and there may be tests you'll want to add in `blocks.test.ts`

src/components/EventControlledContent.tsx
mjoss17: Ok, I see this is what you were mentioning [here](https://github.com/openstax/k12/issues/124#issuecomment-1113603883) about having the math (and tooltip) rendering pushed down to the component leaves.

mjoss17:
Thank you for all the notes @rnathuji. I should have made this a draft, but I understand the scope of this much better now. An end to end playwright test will help expose various holes

mjoss17:
@rnathuji @P-Gill97 This PR has a series of code coverage warnings for single line checks. I'm assuming I should go through and make fixes for these, but I'm curious if there are ever instances in which CodeCov warnings are ignorable?

rnathuji:
@mjoss17 - In this case it doesn't look like codecov is throwing a failed status on the PR, so I think you're talking about the in-line messages it provides about missing coverage. My opinion on these is to use reasonable judgement, keeping in mind that some code is better (or already) covered by e2e tests which aren't captured currently. I'd review them and basically ask yourself:

1. "Is there a useful and meaningful test I can write to cover this code?" - If the answer is "Yes", then go ahead and add it.
2. "Am I writing a brittle / superficial test case just to make the coverage reporter happy?" and / or "Is this code covered by my playwright tests?" - If the answer to either is "Yes", it's probably better to leave it.

Keep in mind that test code, like other code, is something we have to maintain, so we should always be mindful that the value of tests we write is greater than the cost of maintaining them.

As a final note, I think providing feedback on things like this is something we can also include as part of code reviews.

src/components/EventControlledContent.tsx
rnathuji: You should be able to remove all changes to this file in your PR without impacting your test cases (e.g. these changes are all unnecessary given your other code)

src/components/CTABlock.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { EventControlledContent } from './EventControlledContent'
+import { mathifyElement } from '../lib/math'
+
+interface CTABlockProps {
+  contentString: string
+  contentPrompt: string
+  buttonText: string
+  fireEvent?: string
+  waitForEvent?: string
+
```
rnathuji, line 11: Minor: White space nit

src/components/CTABlock.tsx
```diff
@@ -0,0 +1,49 @@
+import { useState, useCallback } from 'react'
+import { EventControlledContent } from './EventControlledContent'
+import { mathifyElement } from '../lib/math'
+
+interface CTABlockProps {
+  contentString: string
+  contentPrompt: string
```
rnathuji, line 7: Minor: `contentString` seems a bit redundant (e.g. including the type in the variable name). Maybe this can just be `content` and then `contentPrompt` --> `prompt` for consistency relative to that?

src/tests/CTABlock.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const renderEvent = new CustomEvent('waitForEvent')
+
+  fireEvent(document, renderEvent)
+
+  screen.queryByText('String')
+  screen.queryByText('Prompt')
+  screen.queryByText('Click me!')
+})
+
+test('CTABlock from parseCTABlock renders on namespaced event', async () => {
```
rnathuji, line 67: It might be nice to include a corresponding `CTABlock from parseCTABlock fires namespaced event on click`

src/lib/blocks.tsx
```diff
@@ -52,6 +57,13 @@ export const blockifyHTML = (htmlContent: string): JSX.Element[] => {
 }
 
 const blockifyElement = (element: HTMLElement): JSX.Element => {
+  if (element.classList.contains(OS_RAISE_IB_CTA_CLASS)) {
+    const maybeCTABlock = parseCTABlock(element)
```
rnathuji, line 61: Minor: If you wanted to get coverage on this  you can extend the `blockifyHTML handles non-block content with interactive blocks` test case similar to what I did in [this PR](https://github.com/openstax/k12-apps-raise/pull/10). (arguably perhaps I should have created a separate test case instead of hijacking that one :man_shrugging: )

src/lib/blocks.tsx
```diff
@@ -93,6 +133,23 @@ const replaceElementWithBlock = (element: HTMLElement, component: JSX.Element):
   )
 }
 
+export const renderCTABlocks = (element: HTMLElement): void => {
```
rnathuji, line 136: Note: Depending on merge order, this should be rewritten in this PR or [this PR](https://github.com/openstax/k12-apps-raise/pull/10) to use the `renderContentBlocksByClass` utility function in the latter.

src/lib/blocks.tsx
```diff
@@ -93,6 +133,23 @@ const replaceElementWithBlock = (element: HTMLElement, component: JSX.Element):
...
+
+  ctaItems.forEach(elem => {
+    const htmlElem = elem as HTMLElement
+    const maybeCTABlock = parseCTABlock(htmlElem)
+
+    if (maybeCTABlock === null) {
+      return
+    }
+
+    htmlElem.innerHTML = ''
```
rnathuji, line 147: This assignment isn't necessary since it's already handled by `replaceElementWithBlock`

src/lib/blocks.tsx
mjoss17: Waiting for your PR to be merged for the time being.

src/components/CTABlock.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (fireEvent !== undefined) {
+      const clickEvent = new CustomEvent(fireEvent)
+      document.dispatchEvent(clickEvent)
+    }
+    setclicked(true)
+  }
+
+  const contentRefCallback = useCallback((node: HTMLDivElement | null): void => {
+    if (node != null) {
+      mathifyElement(node)
```
rnathuji, line 26: Leaving a corresponding note as a marker for [this comment](https://github.com/openstax/k12-apps-raise/pull/8#discussion_r863923597) here.

PR #10, opened 2022-05-02
Implement user input interactive block

rnathuji:
This PR includes an initial implementation that should meet the functional requirements as outlined in the initial `schemas.md`. It also takes a rudimentary stab at some styling, but that will definitely need a better pass once we have UX mockups.

**NOTE:** This PR includes a commit to address [this upstream issue](https://github.com/jaredpalmer/formik/issues/3546) in `formik`'s types that impact React 18. That can be backed out later once we update to a version of the dependency with the type fix.

### DIFF
--- a/src/components/UserInputBlock.tsx
+++ b/src/components/UserInputBlock.tsx
@@ -28,10 +28,6 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent, 
   })
 
   const handleSubmit = async (values: InputFormValues): Promise<void> => {
-    if (values.response === '') {
-      return
-    }
-
     setResponseSubmitted(true)
     if (fireEvent !== undefined) {
       const submitEvent = new CustomEvent(fireEvent)
--- a/src/components/UserInputBlock.tsx
+++ b/src/components/UserInputBlock.tsx
@@ -61,10 +61,11 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent, 
           )}
         </Formik>
       )
+
   const maybeAck = !responseSubmitted
     ? null
     : (
-      <div ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: ack }} />
+      < div ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: ack }} />
       )
 
   return (
--- a/src/components/UserInputBlock.tsx
+++ b/src/components/UserInputBlock.tsx
@@ -4,6 +4,7 @@ import { useState, useCallback } from 'react'
 import { mathifyElement } from '../lib/math'
 
 const DEFAULT_BUTTON_TEXT = 'Submit'
+const DEFAULT_TEXTAREA_ROWS = 3
 
 interface UserInputBlockProps {
   content: string
@@ -54,8 +55,8 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent, 
         >
           {({ isSubmitting }) => (
             <Form>
-              <Field name="response" as="textarea" disabled={isSubmitting}/>
-              <button type="submit" disabled={isSubmitting}>{buttonText !== undefined ? buttonText : DEFAULT_BUTTON_TEXT}</button>
+              <Field name="response" as="textarea" disabled={isSubmitting} rows={DEFAULT_TEXTAREA_ROWS} className="form-control my-3"/>
+              <button type="submit" disabled={isSubmitting} className="btn btn-outline-primary">{buttonText !== undefined ? buttonText : DEFAULT_BUTTON_TEXT}</button>
             </Form>
           )}
         </Formik>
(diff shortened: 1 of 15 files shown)


src/components/UserInputBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+interface InputFormValues {
+  response: string
+}
+
+export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent, buttonText }: UserInputBlockProps): JSX.Element => {
+  const [responseSubmitted, setResponseSubmitted] = useState(false)
+
+  const schema = Yup.object({
+    response: Yup.string().trim().required('Please provide valid input')
+  })
```
mjoss17, line 28: Does this yup response appear in the text box? Or in a popup?

src/styles/bootstrap.scss
```diff
@@ -16,4 +16,61 @@
   @import "bootstrap/scss/helpers";
   @import "bootstrap/scss/buttons";
   @import "bootstrap/scss/forms";
+  @import "bootstrap/scss/utilities";
+  @import "bootstrap/scss/utilities/api";
+
+  // The following are copied from bootstrap.css where the definitions under
+  // :root are otherwise not applied when namespaced so manually defined here
+  --bs-blue: #0d6efd;
```
mjoss17, line 24: Where are all of these used?

src/tests/UserInputBlock.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: '   ' } })
+  })
+
+  screen.getByText('Please provide valid input')
+
+  await act(async () => {
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: 'S' } })
+  })
+  expect(screen.queryByText('Please provide valid input')).toBeNull()
+})
```
mjoss17, line 90: Why do you use this async function to fire the click and change events?

mjoss17 (approved):

src/styles/bootstrap.scss
rnathuji: They are used indirectly by a lot of the Bootstrap utility classes. The `text-danger` I used in styling the error message is what triggered my including this, but I went ahead and put all of the related definitions in here so we hopefully don't have to try and find / diagnose these in the future when things don't work as expected.

src/lib/blocks.tsx
```diff
@@ -93,16 +137,24 @@ const replaceElementWithBlock = (element: HTMLElement, component: JSX.Element):
...
   })
 }
+
+export const renderContentOnlyBlocks = (element: HTMLElement): void => {
+  renderContentBlocksByClass(element, OS_RAISE_IB_CONTENT_CLASS, parseContentOnlyBlock)
+}
+
+export const renderUserInputBlocks = (element: HTMLElement): void => {
+  renderContentBlocksByClass(element, OS_RAISE_IB_INPUT_CLASS, parseUserInputBlock)
+}
```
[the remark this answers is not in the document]
rnathuji: Yep -- when I initially wrote `renderUserInputBlocks` I just felt it was very repetitive to the existing `renderContentOnlyBlocks` so this is a DRY-er implementation that uses a common utility function.

src/components/UserInputBlock.tsx
```diff
@@ -0,0 +1,84 @@
...
+          validationSchema={schema}
+        >
+          {({ isSubmitting }) => (
+            <Form>
+              <Field name="response" as="textarea" disabled={isSubmitting} rows={DEFAULT_TEXTAREA_ROWS} className="form-control my-3"/>
+              <ErrorMessage className="text-danger my-3" component="div" name="response" />
+              <button type="submit" disabled={isSubmitting} className="btn btn-outline-primary">{buttonText !== undefined ? buttonText : DEFAULT_BUTTON_TEXT}</button>
+            </Form>
+          )}
+        </Formik>
```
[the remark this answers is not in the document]
rnathuji: Yeah, `formik` has a lot of magic that it does for you which is super useful and results in doing more with less code, but it may not make sense right away. I'd encourage you to go / read through [their tutorial](https://formik.org/docs/tutorial) just to get a better feel for it since we'll likely be using it in other components (e.g. problem sets) (cc: @P-Gill97 ).

To your question: `isSubmitting` is part of the data bag that `formik` provides (see [doc notes](https://formik.org/docs/api/formik#issubmitting-boolean)). That's being used here as a way to disable the button and input box while submission is happening. For this form there's no network / latency so it's probably impossible to see it, but the case this type of disabling is targeted to address is to not allow the user to accidentally submit a form while it's being processed (e.g. if we were doing a network request and the user clicked the submit button a 100 times for fun or something).

There is a bit of magical "mapping variables to the form fields" which is happening as well (e.g. the `name="response"` is how `formik` ties things to the `InputFormValues` interface I define in this file), but you'll get a better feel for that if you go through the tutorial.

src/tests/UserInputBlock.test.tsx
rnathuji: Ah -- you can try this locally, but if you remove that wrapper you'll get `Warning: An update to Formik inside a test was not wrapped in act(...).`. See related upstream issue [here](https://github.com/jaredpalmer/formik/issues/1543). This was my fix for addressing that in the corresponding test cases.

PR #14, opened 2022-05-13
Create MultiselectProblem

mjoss17:
Draft, some additional test cases may be applicable. Problem with mathjax rendering persists.

### DIFF
--- a/src/components/MultiselectProblem.tsx
+++ b/src/components/MultiselectProblem.tsx
@@ -50,7 +50,7 @@ export const MultiselectProblem = ({
     const options: JSX.Element[] = []
 
     parsedOptionValues.forEach(val => options.push(
-    <div className="form-check">
+    <div key={val} className="form-check">
       <label className="form-check-label" key={val}>
         <Field
         className="form-check-input"
--- a/src/components/MultiselectProblem.tsx
+++ b/src/components/MultiselectProblem.tsx
@@ -28,7 +28,6 @@ export const MultiselectProblem = ({
 
   const contentRefCallback = useCallback((node: HTMLDivElement | null): void => {
     if (node != null) {
-      console.log('mathify')
       mathifyElement(node)
     }
   }, [])
@@ -80,9 +79,6 @@ export const MultiselectProblem = ({
   }
 
   const handleSubmit = async (values: MultiselectFormValues): Promise<void> => {
-    console.log('Handle Submit Called')
-    console.log(values.response)
-    console.log(solutionArray)
     if (compareForm(values.response, solutionArray)) {
       setFeedback(correctResponse)
       solvedCallback()
@@ -99,8 +95,8 @@ export const MultiselectProblem = ({
   }
 
   return (
-    <div className="os-raise-bootstrap">
-      <div className="my-3" ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: content }} />
+    <div className="os-raise-bootstrap" ref={contentRefCallback}>
+      <div className="my-3" dangerouslySetInnerHTML={{ __html: content }} />
       <Formik
         initialValues={{ response: [] }}
         onSubmit={handleSubmit}
--- a/src/components/MultiselectProblem.tsx
+++ b/src/components/MultiselectProblem.tsx
@@ -50,20 +50,19 @@ export const MultiselectProblem = ({
     const options: JSX.Element[] = []
 
     parsedOptionValues.forEach(val => options.push(
-      <>
-    <label className="form-check-label" key={val}>
-      <Field
-      class="form-check-input"
-      type="checkbox"
-      name="response"
-      onChange={(e: React.ChangeEvent<HTMLInputElement>) => { clearFeedback(); setFieldValue('response', modifyModel(values, e)) }}
-      disabled={isSubmitting || formDisabled}
-      value={val}>
-      </Field>
-      {' ' + val}
-    </label>
-    <br></br>
-    </>
+    <div className="form-check">
+      <label className="form-check-label" key={val}>
+        <Field
+        className="form-check-input"
+        type="checkbox"
+        name="response"
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) => { clearFeedback(); setFieldValue('response', modifyModel(values, e)) }}
+        disabled={isSubmitting || formDisabled}
+        value={val}>
+        </Field>
+        {val}
+      </label>
+    </div>
     ))
 
     return options
@@ -99,7 +98,7 @@ export const MultiselectProblem = ({
 
   return (
     <div className="os-raise-bootstrap" ref={contentRefCallback}>
-      <div className="form-check my-3" dangerouslySetInnerHTML={{ __html: content }} />
+      <div className="my-3" dangerouslySetInnerHTML={{ __html: content }} />
       <Formik
         initialValues={{ response: [] }}
         onSubmit={handleSubmit}
@@ -109,7 +108,6 @@ export const MultiselectProblem = ({
           <Form>
             {generateOptions(values, isSubmitting, setFieldValue)}
             <ErrorMessage className="text-danger my-3" component="div" name="response" />
-            <br></br>
             <button type="submit" disabled={isSubmitting || formDisabled} className="btn btn-outline-primary">{buttonText}</button>
             {feedback !== '' ? <div ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: feedback }} className="my-3" /> : null }
           </Form>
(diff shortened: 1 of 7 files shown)


src/tests/MultiselectProblem.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    encourageResponse={''}
+    retryLimit={0}
+    solution={'["Option 2", "Option 3"]'}
+    buttonText={'Check'}
+    />
+  )
+
+  await act(async () => {
+    fireEvent.click(screen.getByLabelText('Option 2'))
+    fireEvent.click(screen.getByLabelText('Option 3'))
```
rnathuji, line 71: You're currently missing coverage that "unclicks" work (the `newSet.delete(e.target.value)` line highlighted by codecov). You might want to create a dedicated test case for this, but you could also just click and then unclick `Option 1` here if you wanted to :man_shrugging: .

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,118 @@
...
+    return Array.from(newSet)
+  }
+
+  const generateOptions = (values: MultiselectFormValues, isSubmitting: boolean, setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void): JSX.Element[] => {
+    const options: JSX.Element[] = []
+
+    parsedOptionValues.forEach(val => options.push(
+    <label key={val}>
+      <Field
+      key={val}
```
rnathuji, line 55: I think you're including `key` values here for React. If so, I don't think you need this since you're rendering with `<label>` as a parent element, and that should be sufficient for React to uniquely identify everything it wants in the list.

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,118 @@
...
+    <div className="os-raise-bootstrap" ref={contentRefCallback}>
+      <div className="my-3" dangerouslySetInnerHTML={{ __html: content }} />
+      <Formik
+        initialValues={{ response: [] }}
+        onSubmit={handleSubmit}
+        validationSchema={schema}
+      >
+        {({ isSubmitting, setFieldValue, values }) => (
+          <Form>
+            <div role="group">
```
rnathuji, line 107: I took a brief look at the docs [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role), but it's not immediately clear to me what this `role="group"` attribute does in practice. Here's what I see in the accessibility with and without it:

image

image

Do we know what the difference is from an accessibility / screen reader standpoint?

(cc: @christopher-mei in case you might have insights from your work)

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,118 @@
...
+      exhaustedCallback()
+      setFeedback(NO_MORE_ATTEMPTS_MESSAGE)
+      setFormDisabled(true)
+    }
+  }
 
-export const MultiselectProblem = ({ solvedCallback, exhaustedCallback, allowedRetryCallback }: MultiSelectProps): JSX.Element => {
-  return <></>
+  return (
+    <div className="os-raise-bootstrap" ref={contentRefCallback}>
```
rnathuji, line 98: Again, we don't want to spend too much time on styling in this iteration, but I think here it'd be nice / easy to incorporate use of the `form-check` and accompanying classes documented [here](https://getbootstrap.com/docs/5.0/forms/checks-radios/) so it looks a bit nicer.

src/lib/math.ts
```diff
@@ -73,13 +73,11 @@ export const loadMathJax = async (): Promise<boolean> => {
 export const mathifyElement = (elem: Element): void => {
   const mathjaxLoaded = window.MathJax !== undefined
   const innerText = elem.textContent
-
```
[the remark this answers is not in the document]
rnathuji: Hmmm...no, definitely still showing up in the PR and on my branch. It may be subtle since it's just the removal of some empty lines:

```
$ git diff main -- src/lib/math.ts
diff --git a/src/lib/math.ts b/src/lib/math.ts
index 161a8f9..369216b 100644
--- a/src/lib/math.ts
+++ b/src/lib/math.ts
@@ -73,13 +73,11 @@ export const loadMathJax = async (): Promise<boolean> => {
 export const mathifyElement = (elem: Element): void => {
   const mathjaxLoaded = window.MathJax !== undefined
   const innerText = elem.textContent
-
   // Only try to mathify if content includes expected delimiters
   // including \( , $$ , or \[
   if (innerText === null || !/\\\(|\$\$|\\\[/.test(innerText)) {
     return
   }
-
   const queueMathJax = (elem: Element): void => {
     window.MathJax.Hub.Queue(['Typeset', window.MathJax.Hub, elem])
   }
```

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+      newSet.delete(e.target.value)
+    }
+    return Array.from(newSet)
+  }
+
+  const generateOptions = (values: MultiselectFormValues, isSubmitting: boolean, setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void): JSX.Element[] => {
+    const options: JSX.Element[] = []
+
+    parsedOptionValues.forEach(val => options.push(
+      <>
```
rnathuji, line 53: instead of empty here, you can do `<div className="form-check">` to get the styling to look right (vs applying the style to the problem below). Also, you'll then want to lift `key={val}` here (it needs to go in the outermost element). You can check / confirm for the errors in the console to be sure these are working properly.

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+    <label className="form-check-label" key={val}>
+      <Field
+      class="form-check-input"
```
rnathuji, line 56: This attribute should be `className` vs `class`

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+      type="checkbox"
+      name="response"
+      onChange={(e: React.ChangeEvent<HTMLInputElement>) => { clearFeedback(); setFieldValue('response', modifyModel(values, e)) }}
+      disabled={isSubmitting || formDisabled}
+      value={val}>
+      </Field>
+      {' ' + val}
```
rnathuji, line 63: Why include the space here? That smells of introducing content for styling and should be avoided.

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+    </label>
+    <br></br>
```
rnathuji, line 65: Same comment as the adding of `' '` above, but I suspect you introduced this for spacing in which case it can be removed. Perhaps you meant to include margins somewhere in which case you should use the bootstrap margin classes.

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+      setFeedback(NO_MORE_ATTEMPTS_MESSAGE)
+      setFormDisabled(true)
+    }
+  }
 
-export const MultiselectProblem = ({ solvedCallback, exhaustedCallback, allowedRetryCallback }: MultiSelectProps): JSX.Element => {
-  return <></>
+  return (
+    <div className="os-raise-bootstrap" ref={contentRefCallback}>
+      <div className="form-check my-3" dangerouslySetInnerHTML={{ __html: content }} />
```
rnathuji, line 102: You don't want to put `form-check` here. It causes the spacing of the problem to be off. See corresponding fix to styling above.

src/components/MultiselectProblem.tsx
```diff
@@ -1,7 +1,120 @@
...
+      <Formik
+        initialValues={{ response: [] }}
+        onSubmit={handleSubmit}
+        validationSchema={schema}
+      >
+        {({ isSubmitting, setFieldValue, values }) => (
+          <Form>
+            {generateOptions(values, isSubmitting, setFieldValue)}
+            <ErrorMessage className="text-danger my-3" component="div" name="response" />
+            <br></br>
```
rnathuji, line 112: Same comment above about not introducing `<br>` elements to create spacing. LMK and we can chat more about the use of styling / CSS.

mjoss17: Ah. I see what you mean, I got carried away after looking at the "form-check" additions making the vite app look nicer.

src/components/MultiselectProblem.tsx
rnathuji: @mjoss17 - I think you may have missed the second half of this comment about moving `key={val}` to the new `<div className="form-check">` line you added. Otherwise right now you still get the following React gripe:

image


If you can fix :point_up: real quick I think everything else looks great!

mjoss17: Thank you, Ah!

PR #15, opened 2022-05-16
Create user input problem 

P-Gill97:


src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+    <Formik
+          initialValues={{ response: '' }}
+          onSubmit={handleSubmit}
+          validationSchema={schema}
+        >
+          {({ isSubmitting, setFieldValue }) => (
+            <Form >
+              <Field
+              name="response"
+              as="textarea"
```
rnathuji, line 80: I think for problem sets we can use standard `<input>` element instead of the multi-line `<textarea>`. That's the default so you could just remove this line to go that route.

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+              disabled={inputDisabled}
+              onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => { clearFeedback(); setFieldValue('response', e.target.value) }}
+              className="form-control my-3" />
```
rnathuji, line 83: If you wanted to spruce up the styling a bit so the input is not `full-width` per [here](https://github.com/twbs/bootstrap/blob/416f4905f13c2ca3096343e0a5a4934c8c0ced3f/scss/forms/_form-control.scss#L7), you can define / use a `.os-form-control` class in our `bootstrap.scss` as:

```css
  .os-form-control {
    @extend .form-control;

    width: unset;
  }
```

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+          initialValues={{ response: '' }}
+          onSubmit={handleSubmit}
+          validationSchema={schema}
+        >
+          {({ isSubmitting, setFieldValue }) => (
+            <Form >
+              <Field
+              name="response"
+              as="textarea"
+              disabled={inputDisabled}
```
rnathuji, line 81: Since our submit handlers are pretty close to instantaneous it's hard to see this potential issue, but it might be good to include `isSubmitting` on the `disabled` attribute to avoid "double clicks" while the form is being processed (e.g. something like `disabled={isSubmitting || formDisabled}`)

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+              onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => { clearFeedback(); setFieldValue('response', e.target.value) }}
+              className="form-control my-3" />
+              <ErrorMessage className="text-danger my-3" component="div" name="response" />
+              <button type="submit" disabled={inputDisabled} className="btn btn-outline-primary">{buttonText}</button>
```
rnathuji, line 85: Same comment on the `disabled` attribute here as above.

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+  solution, retryLimit, content, comparator, encourageResponse, buttonText, correctResponse
+}: InputProblemProps): JSX.Element => {
+  const [retriesAllowed, setRetriesAllowed] = useState(0)
+  const [inputDisabled, setInputDisabled] = useState(false)
+  const [feedback, setFeedback] = useState('')
+
+  const schema = (): Yup.SchemaOf<InputSchema> => {
+    if (comparator === 'integer') {
+      return Yup.object({
+        response: Yup.number().integer().typeError('Please provide an Integer').required('Please provide an Integer')
```
rnathuji, line 30: I found this randomly, but it seems if you provide an input like `24.5` you get the message `response must be an integer`. I think that's the default for the `integer()` here, so if you wanted to put a custom message you can make that `integer("Please provide an Integer not a float")` or something like that.

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+  }
+
+  const contentRefCallback = useCallback((node: HTMLDivElement | null): void => {
+    if (node != null) {
+      mathifyElement(node)
+    }
+  }, [])
+
+  const handleSubmit = async (values: InputFormValues): Promise<void> => {
+    if (values.response.trim() === solution.trim()) {
```
rnathuji, line 50: I don't think this comparison is sufficient, and we probably need one that is a function of the actual `comparator`. For example, if I define the problem as:

```html
<div class="os-raise-ib-pset-problem" data-problem-type="input" data-solution="42.0" data-problem-comparator="float">
```

I should be able to provide `42` or `42.0` or `42.0000` and get it correct. 

I would suggest:

* For `text` use `===` after converting both to lower case (e.g. so `red` matches a solution definition of `Red`)
* For `int` compare using results from `parseInt` 
* For `float` compare using results from `parseFloat`

src/tests/InputProblem.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+          />
+  )
+  await act(async () => {
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: 'Apple ' } })
+    screen.getByRole('button').click()
+  })
+  await screen.findByText('Correct!')
+  expect(solvedHandler).toBeCalledTimes(1)
+  expect(exhaustedHandler).toBeCalledTimes(0)
+  expect(allowedRetryHandler).toBeCalledTimes(0)
```
rnathuji, line 52: Can we add assertions that the form is disabled when answered correctly to this test case?

src/tests/InputProblem.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  await act(async () => {
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: '3' } })
+    screen.getByRole('button').click()
+  })
+
+  await screen.findByText('No more attempts allowed')
+  expect(solvedHandler).toBeCalledTimes(0)
+  expect(exhaustedHandler).toBeCalledTimes(1)
+  expect(allowedRetryHandler).toBeCalledTimes(3)
```
rnathuji, line 183: Can we add assertions that the form is disabled when the form is exhausted to this test case?

src/components/InputProblem.tsx
```diff
@@ -1,7 +1,92 @@
...
+      setFeedback(correctResponse)
+      solvedCallback()
+      setInputDisabled(true)
+    } else if (retryLimit === 0 || retriesAllowed !== retryLimit) {
+      setRetriesAllowed(currRetries => currRetries + 1)
+      setFeedback(encourageResponse)
+      allowedRetryCallback()
+    } else {
+      exhaustedCallback()
+      setFeedback('No more attempts allowed')
```
rnathuji, line 60: Can you incorporate use of [this shared constant](https://github.com/openstax/k12-apps-raise/blob/6c7c3d02a0e25fb0e6800800c87c737389a5d300/src/components/ProblemSetBlock.tsx#L31) for the string here?

src/components/InputProblem.tsx
rnathuji: @P-Gill97 - Everything looks good to me on this PR except for 2 UI related items:

1. Can we add the style definition for `.os-form-control` in my previous comment [after this line](https://github.com/openstax/k12-apps-raise/blob/6c7c3d02a0e25fb0e6800800c87c737389a5d300/src/styles/bootstrap.scss#L83) in `bootstrap.scss`, and then apply the class here so it's `className="os-form-control my-3"`? That will give us the nicer Bootstrap styling on the input box versus the default browser.

2. Can we move the `ErrorMessage` on top of the `<button>` to be consistent with the other components? So it's something like:

```javascript
<ErrorMessage className="text-danger mb-3" component="div" name="response" />
<button type="submit" disabled={inputDisabled || isSubmitting} className="btn btn-outline-primary ml-3">{buttonText}</button>
```

rnathuji: @P-Gill97 - Just saw your last commit and while the styling looks good now I think the element ordering is still not quite right as now the feedback is above the button. We want to be consistent with other components such as [here](https://github.com/openstax/k12-apps-raise/blob/6c7c3d02a0e25fb0e6800800c87c737389a5d300/src/components/MultiselectProblem.tsx#L110-L112) where it's:

```javascript
<ErrorMessage />
<button />
{feedback }
```

Unfortunately this consistency is by convention right now so we have to keep an eye out for it manually. I already have it on my list to create an issue to refactor some of this problem type code which will be an opportunity to codify the conventions in a single location.

P-Gill97: Sounds good. Fixed it.

rnathuji (approved):

PR #16, opened 2022-05-18
Add MultipleChoice question type

mjoss17:


specs/blocks.spec.ts
```diff
@@ -145,3 +145,31 @@ test('math is rendered in DropdownProblem question', async ({ page }) => {
   await page.locator('text=Check').click()
   await page.waitForSelector('#correct .MathJax')
 })
+
+test('math is rendered in MultipleChoiceProblem question', async ({ page }) => {
```
rnathuji, line 149: Since one of the potential use cases for `multiplechoice` over `dropdown` is to support questions with math choices, I found a little quirk which is worth discussing. First off, we should add a test case like your existing `test('math is rendered in Multiselect answer')` case where the answers are math:

```javascript
test('math is rendered in Multiplechoice answer and can be selected as answer', async ({ page }) => {
}
```

What you'll  notice is there's a problem if you do:

```javascript
test('math is rendered in Multiplechoice answer', async ({ page }) => {
  const htmlContent = `
  <div class="os-raise-ib-pset" data-schema-version="1.0">
    <div class="os-raise-ib-pset-problem" data-problem-type="multiplechoice" data-solution='\\\\(x^2\\\\)' data-solution-options='["\\\\(x^2\\\\)", "blue", "green"]'>
```

The problem is not solved as correct if you answer `x^2`. The reason is that JSON escaping requires double `\\` but in the current schema you have `data-solution` is just a string. There are two options:

1. Keep things this way, and content developers will need to use `\( \)` in the solution but `\\( \\)` in the JSON per:

```html
<div class="os-raise-ib-pset-problem" data-problem-type="multiplechoice" data-solution="\( x^2 \)" data-solution-options='["\\( x^2 \\)", "\\( x+2 \\)", "\\( 2*x \\)"]'>
```

The downside of :point_up: is it will be different than `multiselect` where both are JSON so it looks like:

```html
<div class="os-raise-ib-pset-problem" data-problem-type="multiselect" data-solution='["\\( x^2 \\)", "\\( x+2 \\)"]' data-solution-options='["\\( x^2 \\)", "\\( x+2 \\)", "\\( 2*x \\)"]'>
```

We could make `multiplechoice` `data-solution` JSON as well I guess, just expecting an array of size 1, but that's also a bit confusing. Also then it diverges from `dropdown` in terms of the schema for non-array solutions.

So what's the lesser evil? :thinking:  Right now I'd vote for using the approach where content developers use `data-solution="\( x^2 \)" ` and just document + hope they actually test things out when building the problem :joy: , but would like to hear what you think.

specs/blocks.spec.ts
```diff
@@ -145,3 +145,31 @@ test('math is rendered in DropdownProblem question', async ({ page }) => {
...
+  const htmlContent = `
+  <div class="os-raise-ib-pset" data-schema-version="1.0">
```
rnathuji, line 151: I wasn't sure where to put this general comment, so shoving it here: As part of this PR let's make a couple of updates to `schema.md`:

1. Include details for the new question type
2. Specifically call out how to use MathJax math for `multiselect` and `multiplechoice` answer options + MathJax math rendering is not supported for `dropdown`

src/components/ProblemSetBlock.tsx
```diff
@@ -32,6 +33,7 @@ export const NO_MORE_ATTEMPTS_MESSAGE = 'No more attempts allowed'
 export const PROBLEM_TYPE_INPUT = 'input'
 export const PROBLEM_TYPE_DROPDOWN = 'dropdown'
 export const PROBLEM_TYPE_MULTISELECT = 'multiselect'
+export const PROBLEM_TYPE_MULTIPLECHOICE = 'multiplechoice'
```
rnathuji, line 36: Can we include a check similar to what's there for multiselect and dropdown [here](https://github.com/openstax/k12-apps-raise/blob/6c7c3d02a0e25fb0e6800800c87c737389a5d300/src/lib/blocks.tsx#L220) where we validate `data-solution-options` is provided if the problem type is `PROBLEM_TYPE_MULTIPLECHOICE`?

src/components/MultipleChoiceProblem.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const generateOptions = (values: MultipleChoiceFormValues, isSubmitting: boolean, setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void): JSX.Element[] => {
+    const options: JSX.Element[] = []
+
+    parsedOptionValues.forEach(val => options.push(
+    <div key={val} className="form-check">
+      <label className="form-check-label">
+        <Field
+        className="form-radio-input"
```
rnathuji, line 45: Based on [this documentation](https://getbootstrap.com/docs/5.2/forms/checks-radios/#radios) it looks like the correct class is `form-check-input` here (that might also explain / address the odd rendering otherwise).

rnathuji (commented):
@mjoss17 - This PR looks really good! I have a few minor comments and one that might merit some team discussion during our engineering check-in tomorrow.

specs/blocks.spec.ts
mjoss17: Sounds good 👍 I'll append the new question type details, but I'd love to discuss the MathJax details from above at the engineering meeting too.

specs/blocks.spec.ts
rnathuji: @P-Gill97 - Just a `cc:` here so you have a chance to review prior to our check-in. Here are my current thoughts:

- It'd be nice to be able to document `multiplechoice` as being the exact same schema as `dropdown`, just with a different value for `data-problem-type`.
- If we do :point_up: , we probably can just add a specific :warning:  How to use math in answer options for `multiplechoice` and `multiselect` :warning:  section to the documentation which calls out the need to escape `\` in anything that is interpreted as JSON and provide examples.

mjoss17: This is interesting. I guess my question is, is there not some way we can consistently alter the equivalency  conditional, like using html.escape, or some sort of string parser to alter the solution? I figured that would be a nice way to keep the schemas consistent for content developers, but I understand if it brings too much "string altering" complexity and potential test cases into the mix.

rnathuji: Well, the thought I had around encoding the answer as a JSON array of 1 for `multiplechoice` is basically doing that, but while that's technically feasible I think it introduces additional issues / confusion that probably just aren't worth it. I definitely don't want to get into the game of trying to "hack" unescaping because that is just asking for unintended consequences. I think it's safest + most reasonable to just say: If it's JSON, escape like you would for JSON, and when it's not, don't.

rnathuji: Throwing tangible examples of specifying math answers for `multiselect` and `multiplechoice` into this thread for reference during discussion:

```html
  <div class="os-raise-ib-pset-problem" data-problem-type="multiplechoice" data-solution="\( x^2 \)" data-solution-options='["\\( x^2 \\)", "\\( x+2 \\)", "\\( \\frac{2*x}{y} \\)"]'>
    <div class="os-raise-ib-pset-problem-content">
      <p>Multiple choice problem content: \( x^2 \)</p>
    </div>
  </div>
  <div class="os-raise-ib-pset-problem" data-problem-type="multiselect" data-solution='["\\( x^2 \\)", "\\( x+2 \\)"]' data-solution-options='["\\( x^2 \\)", "\\( x+2 \\)", "\\( 2*x \\)"]'>
    <div class="os-raise-ib-pset-problem-content">
      <p>Multiselect problem content: \( x^2 \)</p>
    </div>
  </div>
```

mjoss17: I was thinking, if math is something that is continuously being added to content in different places, it could be interesting to have a quick tool that generates math for specific problems, so that content developers and researchers can use a common math language. I'm not sure where this would fit in an editor yet but something that spews LaTeX to escaped strings - or LaTeX to these 'double escaped strings' would be extremely helpful for them. Maybe this exists behind the scenes in our VSCode Editor...?

rnathuji: Oh, I see. I don't think that's something we would build for a couple of reasons:

* In general, I think (perhaps incorrectly) folks used to developing math content are pretty familiar with LaTeX. The bigger gap is usually if we ask them to do MathML like in our books. I believe there folks sometimes use tools that generate MathML from LaTeX, but AFAIK even in POET we don't build that in.
* If we're trying to address users that aren't familiar with LaTeX, it's difficult to know what the "right" baseline is since that's as close to a "typable" standard that exists. Funny enough, if you ever dig into the guts of things like [pandoc](https://pandoc.org/), as part of handling different math content LaTeX is the canonical intermediary format (e.g. convert `.docx` math into LaTeX and then LaTeX into whatever the new document type needs). I don't think we need / want to reinvent any of these wheels.

specs/blocks.spec.ts
rnathuji: @mjoss17 - Actually, it occurred to me just now that you can ignore my comment about adding `schema.md` changes to this PR. Moving forward let's spin up PRs to update the documentation **after** the code hits production. That way WW is not getting confused if they see / try things before the code ships.

PR #20, opened 2022-05-26
Add attempts exhausted override

mjoss17:
This commit contains an issue in which attempts-exhaused responses don't render math correctly

### DIFF
--- a/src/components/DropdownProblem.tsx
+++ b/src/components/DropdownProblem.tsx
@@ -1,4 +1,4 @@
-import { BaseProblemProps, NO_MORE_ATTEMPTS_MESSAGE } from './ProblemSetBlock'
+import { BaseProblemProps } from './ProblemSetBlock'
 import { determineFeedback } from '../lib/problems'
 import { Formik, Form, Field, ErrorMessage } from 'formik'
 import { mathifyElement } from '../lib/math'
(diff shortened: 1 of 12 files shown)


src/components/DropdownProblem.tsx
```diff
@@ -59,8 +59,8 @@ export const DropdownProblem = ({
       setFeedback(determineFeedback(values.response, encourageResponse, answerResponses, evaluateInput))
       allowedRetryCallback()
     } else {
+      setFeedback(attemptsExhaustedResponse)
```
rnathuji, line 62: Minor clarification question: Is there a specific reason why the relative ordering of this invocation was moved to come before `exhaustedCallback()`, or just happenstance during development?

rnathuji (approved):

src/components/DropdownProblem.tsx
mjoss17: Ah - this was me messing around to see if I could fix the callback before we found that bug Friday afternoon. I'm happy to move it back to match the others, couldn't find any functional difference with this change.

rnathuji: I don't think it matters much either way. I just wanted to be sure there wasn't something you found about the ordering of these mattering :sweat_smile:

PR #22, opened 2022-06-02
Render spinner / errors in content loader component

rnathuji:


src/components/ContentLoader.tsx
```diff
@@ -37,7 +42,7 @@ export const ContentLoader = ({ contentId }: ContentLoaderProps): JSX.Element =>
   }
 
   useEffect(() => {
-    fetchContent().catch(() => {})
+    fetchContent().catch(() => { })
```
mjoss17, line 45: Confused what the difference is with this change?

PR #23, opened 2022-06-13
Add desmos block

P-Gill97:


src/tests/DesmosBlock.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+  loadDesmos: jest.fn()
+}))
+
+test('Desmos block calls script', async () => {
+  render(
+      <DesmosBlock width={'500'} height={'500'} equations={'["x=5", "(1,2)"]'} expressions={'true'} scaleTop={'10'} scaleBottom={'-10'} scaleLeft={'-10'} scaleRight={'10'} />
+  )
+
+  expect(loadDesmos).toBeCalled()
+})
```
P-Gill97, line 16: @rnathuji Kind of stumped on this one. How can I mock a function like  Desmos.GraphingCalculator if I can't explicitly import the Desmos library since it is a script. 🤔 Have you ran into something like this before? 
I want to be able to test something like this 

`expect(Desmos.GraphingCalculator).toBeCalledWith(args)`

rnathuji: :thinking: -- Ah, I see. So since the script is going to set a global variable, you can mock things on `window` like the following:

```javascript
jest.mock('../lib/desmos.ts', () => ({
  loadDesmos: jest.fn(async () => {})
}))

test('Desmos block calls script', async () => {
  const setExpression = jest.fn()
  const setMathBounds = jest.fn()
  const gc = jest.fn(() => {
    return {
      setExpression: setExpression,
      setMathBounds: setMathBounds
    }
  })
  window.Desmos = {
    GraphingCalculator: gc
  } as any

  render(
      <DesmosBlock width={'500'} height={'500'} equations={'["x=5", "(1,2)"]'} expressions={'true'} scaleTop={'10'} scaleBottom={'-10'} scaleLeft={'-10'} scaleRight={'10'} />
  )

  expect(loadDesmos).toBeCalled()

  await waitFor(() => expect(gc).toBeCalled())
  expect(setExpression).toBeCalled()
  expect(setMathBounds).toBeCalled()
})
```

Note: In the example I updated the mock for `loadDesmos` to use a different signature for `loadDesmos` as I don't think you really need to pass a function to work that way. Since it's async, you can just leave it with no inputs and use it as follows:

```javascript
  useEffect(() => {
    if (desmosLoaded) {
      renderCalculator()
      return
    }
    loadDesmos().then(() => setDesmosLoaded(true)).catch(() => {})
  }, [desmosLoaded])
```

P-Gill97: Got it! Thanks!

P-Gill97: > Note: In the example I updated the mock for `loadDesmos` to use a different signature for `loadDesmos` as I don't think you really need to pass a function to work that way. Since it's async, you can just leave it with no inputs and use it as follows:

Using this new signature breaks the code and only one instance of calculator will show up in case there are multiple desmos blocks. I am trying to figure out what I need to change in `lib/desmos.ts` to fix this. 🤔

rnathuji: @P-Gill97 - Interesting. Perhaps we can pair for a bit / :parking: this for engineering check-in tomorrow.

P-Gill97:
Codecov complain about desmos.ts but it is tested in playwright tests.

src/tests/DesmosBlock.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+  expect(loadDesmos).toBeCalled()
+
+  await waitFor(() => expect(graphingCalculator).not.toBeCalled())
+  expect(setExpression).not.toBeCalled()
+  expect(setMathBounds).not.toBeCalled()
+})
+
+test('DesmosBlock calculator does render if waitForEvent does not fire', async () => {
```
rnathuji, line 58: Super minor: I noticed that this says "does not fire" but I think it's actually "does fire"

PR #25, opened 2022-06-27
Disable input textbox and button. 

P-Gill97:


src/components/UserInputBlock.tsx
```diff
@@ -48,23 +48,21 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
               <ErrorMessage className="text-danger my-3" component="div" name="response" />
-              <button type="submit" disabled={isSubmitting} className="btn btn-outline-primary">{buttonText}</button>
+              <button type="submit" disabled={isSubmitting || responseSubmitted} className="btn btn-outline-primary">{buttonText}</button>
             </Form>
           )}
         </Formik>
-      )
+  )
 
   const maybeAck = !responseSubmitted
```
rnathuji, line 67: Minor: Since the form is always rendered, as a minimal UI tweak what do you think about adding a bit of margin to the ack message so it's not up against the button? Something like:

```javascript
< div className="my-3" ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: ack }} />
```

rnathuji (approved):

PR #26, opened 2022-06-30
Add work around for buggy moodle mathjax filter behavior

rnathuji:
The Moodle mathjax filter parses all stored content as a string in
the PHP backend and transforms any content it finds with math
delimiters. Unfortunately this ends up impacting all content including
HTML comments, attributes, etc., resulting content preview errors for
content devs that write valid HTML for our interactive blocks. This
change includes a runtime "patch" that basically tries to undo the
damage that the Moodle filter plugin does on the backend.


mjoss17 (approved):
This works / passes test cases well for me

PR #40, opened 2022-09-16
Create max input for InputProblem and UserInputBlock

mjoss17:


src/tests/UserInputBlock.test.tsx
```diff
@@ -88,6 +88,33 @@ test('UserInputBlock requires non-empty input', async () => {
...
+  })
+
+  screen.getByText('Content text')
+  screen.getByText('Prompt text')
+  screen.getByRole('textbox')
+  screen.getByRole('button')
+  screen.getByText('Please provide valid input')
+
+  await act(async () => {
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: new Array(MAX_CHARACTER_INPUT_BLOCK_LENGTH + 2).join('a') } })
```
rnathuji, line 107: Out of curiosity: Is there a specific reason why this test case uses a different idiom from the others where you employ `String.repeat()`?

src/components/InputProblem.tsx
```diff
@@ -5,6 +5,8 @@ import { Formik, Form, Field, ErrorMessage } from 'formik'
 import { mathifyElement } from '../lib/math'
 import * as Yup from 'yup'
 
+export const MAX_CHARACTER_INPUT_PROBLEM_LENGTH = 10
```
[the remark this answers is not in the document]
mjoss17: This actually, I'm realizing now, is indicative of a problem - wherein when the MAX_CHARACTER_LENGTH is set past 20 digits, the flag is never thrown. Sometimes, in the jest tests, the input is treated as a string, but in my example courses - no flag is even thrown. Still looking into this. (I was seeing this Friday and never got to the bottom of it)

src/tests/UserInputBlock.test.tsx
mjoss17: Nope - Just a typo, I'll change to the other version. I think it's cleaner

rnathuji:
@mjoss17 - Ah, I see. I wonder if maybe this is inherently due to the use of `mixed.test` with numbers. Does a length restriction even make sense on those comparator types? Javascript is going to convert to an `int` / `float` which will be bounded, so I'm thinking the root issue here is we really only need length restrictions on text input. Sorry, I should have thought about that more in the initial review.

src/components/InputProblem.tsx
```diff
@@ -28,16 +30,24 @@ export const InputProblem = ({
   const schema = (): Yup.SchemaOf<InputSchema> => {
     if (comparator.toLowerCase() === 'integer') {
       return Yup.object({
-        response: Yup.number().integer('Please provide an Integer').typeError('Please provide an Integer').required('Please provide an Integer')
+        response: Yup.number()
+          .integer('Please provide an Integer')
+          .typeError('Please provide an Integer')
+          .required('Please provide an Integer')
```
rnathuji, line 36: Both to keep this PR clean and to minimize potential merge conflicts with the PR @P-Gill97 is working on, I'd revert this altogether so no diff appears.

src/components/InputProblem.tsx
```diff
@@ -28,16 +30,24 @@ export const InputProblem = ({
...
       })
     }
     if (comparator.toLowerCase() === 'float') {
       return Yup.object({
-        response: Yup.number().typeError('Please provide an number').required('Please provide a number')
+        response: Yup.number()
+          .typeError('Please provide a number')
+          .required('Please provide a number')
```
rnathuji, line 43: Both to keep this PR clean and to minimize potential merge conflicts with the PR @P-Gill97 is working on, I'd revert this altogether so no diff appears.

src/components/InputProblem.tsx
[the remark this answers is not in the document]
rnathuji: Nudging this comment. I think you changed it from `10` to `50` instead of `500`.

rnathuji (approved):

PR #42, opened 2022-09-19
Modify ContentLoader to serve variants

mjoss17:


rnathuji:
@mjoss17 - Just FYI: Since we don't want this spike code to land on `main`, I created a `k12-221/spike-code` branch with the latest from `main` and updated this PR to merge against that.

mjoss17:
@rnathuji Do you understand the Playwright error that I'm receiving here?

rnathuji:
@mjoss17 - Ah, this is the same issue I hit in the past [here](https://github.com/openstax/k12-apps-raise/commit/0d871df7fdade1cad0df8139b89a497be5f90419). One way to fix that is with the following change:

```patch
diff --git a/src/lib/utils.ts b/src/lib/utils.ts
index e86c44b..2770eb6 100644
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,5 +1,5 @@
 import * as variantMappings from '../../data/variant-mapping.json'
-import { ContentVariant } from '../components/ContentLoader'
+import type { ContentVariant } from '../components/ContentLoader'
 
 export const loadScriptTag = async (srcValue: string): Promise<void> => {
   const head = document.getElementsByTagName('head')[0]
```

src/lib/utils.ts
```diff
@@ -34,3 +37,47 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
...
+  return {
+    userId,
+    courseId,
+    instance
+  }
+}
+
+function getVariantMappings(host: string, courseid: string): string {
+  const mappingData = variantMappings as any
+  const maybeDataMapping = mappingData.default[host]?.courses[courseid]
```
rnathuji, line 57: Just curious: What's the `.default` for?

src/lib/utils.ts
```diff
@@ -34,3 +37,47 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
 
   await loadedPromise
 }
+
+export const getUserVariables = (): {userId: string, courseId: string, instance: string} | null => {
+  const userId = '2'
```
rnathuji, line 42: We don't use / need `userId` right now so I'd remove it to avoid confusion. 

Also, what do you think about calling this `getUserContext` or `getCurrentContext`?

src/lib/utils.ts
```diff
@@ -34,3 +37,47 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
 
   await loadedPromise
 }
+
+export const getUserVariables = (): {userId: string, courseId: string, instance: string} | null => {
```
rnathuji, line 41: It might be nice to have this always return an object and then let `getVariantMapings` decide what it can / can't live without. Basically instead us a signature like:

```javascript
getUserVariables = (): {courseId: string | undefined, instance: string | undefined} => {
```

Then you can always just call `getVariantMappings` (this avoids the need to initialize `varName` + declare the implied default variant name in two places) and instead have `getVariantMappings` be something like:

```javascript
function getVariantMappings(host: string | undefined, courseId: string | undefined): string {
  const defaultVariant = 'main'

  if (courseId === undefined || host === undefined) {
    return defaultVariant
  }

  const mappingData = variantMappings as any
  const maybeDataMapping = mappingData.default[host]?.courses[courseId]
  if (maybeDataMapping !== undefined) {
    return maybeDataMapping
  }

  return defaultVariant
}
```

src/lib/utils.ts
```diff
@@ -34,3 +37,47 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
...
+  let html
+  variants.forEach(item => {
+    if (item.variant === varName) {
+      html = item.html
+    }
+  })
+  if (html === undefined) {
+    html = variants[0].html // return whatever variant exists
+  }
+  return html
```
rnathuji, line 82: In so much as it might be useful for debug purposes to render an error if we unexpectedly find no matching variatn, what do you think about changing this function signature a bit?:

```javascript
export const getVariant = (variants: ContentVariant[]): string | undefined => {
  const userValues = getUserVariables()

  const varName = getVariantMappings(userValues.instance, userValues.courseId)

  const maybeMatch = variants.find(item => item.variant === varName)
  if (maybeMatch === undefined) {
    return maybeMatch
  } else {
    return maybeMatch.html
  }
}
```

With :point_up: , `ContentLoader.tsx` can then display an error:

```javascript
      const data = await response.json() as ContentResponse
      const htmlContent = getVariant(data.content)

      if (htmlContent === undefined) {
        setFetchStatus(FetchStatus.FetchFailure)
        return
      }
```

src/components/ContentLoader.tsx
```diff
@@ -32,7 +38,7 @@ export const ContentLoader = ({ contentId }: ContentLoaderProps): JSX.Element =>
       }
 
       const data = await response.json() as ContentResponse
-      const htmlContent = data.content[0].html
+      const htmlContent = getVariant(data.content)
```
rnathuji, line 41: See related comment in `utils.ts`, but it'd be nice to set an error if we somehow don't find an expected variant in the context (either `main` or the one defined in `variant-mapping.json`)

src/lib/utils.ts
```diff
@@ -34,3 +37,47 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
...
+    return null
+  }
+  return {
+    userId,
+    courseId,
+    instance
+  }
+}
+
+function getVariantMappings(host: string, courseid: string): string {
```
rnathuji, line 55: There's a little bit of inconsistency where you call it `host` here and `instance` in the object returned by `getUserVariables`. I personally like `host` here, so maybe we can modify the other ones just for consistency?

src/lib/utils.ts
mjoss17: It was just the way the object was structured when imported in this manner - the whole thing was wrapped in a "default" field. I couldn't find a way to create the object otherwise. 🤷‍♂️ (sry not a fun answer)

rnathuji: Oh -- I see. I think to avoid this you can do your import as:

```javascript
import variantMappings from '../../data/variant-mapping.json'
```

Then this code just becomes:


```javascript
  const mappingData = variantMappings as any
  const maybeDataMapping = mappingData[host]?.courses[courseid]
```

mjoss17: Ahhhh 👏 nice

mjoss17:
@rnathuji Would it be helpful for me to push a branch of k12-contents-repo that offers an easy way to test this variant delivery code? If you haven't already created one...

rnathuji:
@mjoss17 - Ah, we are thinking the same today :joy: . I just picked up [k12-225](https://github.com/openstax/k12/issues/225) and setup `k12-221/spike-content` branch a couple of minutes ago. I'm going to create a variant page for Unit 2 O&R and push that up to my sandbox demo so we can demo per that issue description.

rnathuji (approved):

PR #47, opened 2022-10-07
Implement ContentLoader to serve variants

mjoss17:


mjoss17:
@rnathuji I'm having trouble still on the syntax for mocking `getVariant()` in ContentLoader.test.tsx. I feel like it must be something silly I'm missing... Wondering if you see it immediately.

rnathuji:
@mjoss17 - IIRC mocking inside of the test case won't work. You have to do it at the file level like the mock in the same test file for `env.ts`.

mjoss17:
@rnathuji If we mock `geVariant()` completely in `ContentLoader.test.ts` There really is no difference between loading a variant and loading the default content. Each just accesses the JSON with both variants - sends both to the `getVariant()` mock, and receives back a whatever the mock gives them. 

I don't feel like making the mock capable of returning different things really exercises that much more of the ContentLoader... I think having this new test and the previous 'get main variant' test is redundant

rnathuji:
@mjoss17 - I see what you're saying and that makes sense to me. I think the playwright test is exercising everything anyways, so sounds good as far as leaving the `ContentLoader` unit tests alone 👍 .

src/lib/utils.ts
```diff
@@ -34,3 +37,40 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
...
+  const host = window.location.host
+  return {
+    courseId,
+    host
+  }
+}
+
+function getVariantMappings(host: string | undefined, courseId: string | undefined): string {
+  const defaultVariant = 'main'
+  console.log(courseId, host)
```
rnathuji, line 52: Minor: Remove this debug line.

specs/variants.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/test'
+import { mockVariantContentRequest } from './utils'
+
+test('user input block is rendered with math and tooltip', async ({ page }) => {
```
rnathuji, line 4: This test case name needs to be corrected to reflect the test.

src/tests/ContentLoader.test.tsx
```diff
@@ -25,6 +25,10 @@ jest.mock('../lib/env.ts', () => ({
   }
 }))
 
+jest.mock('../lib/utils', () => ({
+  getVariant: jest.fn(x => '<p>Variant content</p>')
+}))
+
```
rnathuji, line 31: Per [this comment](https://github.com/openstax/k12-apps-raise/pull/47#issuecomment-1272109242) we can remove this, and it's the reason for why the CI is currently failing. We get essentially the same coverage if you leave the existing tests but add the `variant: 'main'` to the content returned by the mock on line 11 above.

src/tests/ContentLoader.test.tsx
```diff
@@ -36,6 +40,13 @@ test('content is fetched and rendered on success', async () => {
   await screen.findByText('Test content')
 })
 
+test('variant content is fetched and rendered on success', async () => {
+  render(
+    <ContentLoader contentId='test-content'/>
+  )
+  await screen.findByText('Variant content')
+})
+
```
rnathuji, line 49: Per [your comment](https://github.com/openstax/k12-apps-raise/pull/47#issuecomment-1272088665) this test case doesn't add any coverage / value, so we can rely on the playwright test for this change.

src/lib/utils.ts
```diff
@@ -34,3 +37,40 @@ export const loadScriptTag = async (srcValue: string): Promise<void> => {
...
+export const getCurrentContext = (): {courseId: string | undefined, host: string | undefined} => {
+  const courseId = window.M?.cfg.courseId
+  const host = window.location.host
+  return {
+    courseId,
+    host
+  }
+}
+
+function getVariantMappings(host: string | undefined, courseId: string | undefined): string {
```
rnathuji, line 50: Minor: Would calling this `getVariantMapping` be more accurate since it is a singular mapping value returned?

src/lib/utils.ts
```diff
@@ -1,3 +1,6 @@
+import variantMappings from '../../data/variant-mapping.json'
```
rnathuji, line 1: Minor: It'd be nice to be consistent here on the naming in terms of singular vs plural. E.g one of:

* `variantMappings` and `variant-mappings.json` 
* `variantMapping` and `variant-mapping.json`

I'd maybe lean towards the former and then `getVariantMapping` as the name for the function which identifies the appropriate mapping from `variantMappings` :man_shrugging: .

rnathuji (commented):
I wasn't sure if this was ready for review as the CI is failing, but it was short + in the `Needs Review` column so I did a quick pass.

mjoss17:
@rnathuji Are these CI / CodeCov errors thrown because codecov isn't analyzing playwright tests? I believe I remember us talking about something like that.

rnathuji:
@mjoss17 - Yep, in fact we opened [k12-130](https://github.com/openstax/k12/issues/130) for that. The repo is setup so it doesn't require codecov to pass, though, just the unit tests and an approval, so you can ignore that in this PR.

data/variant-mappings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "localhost:3001": {
+        "courses": {
+            "1": "main",
```
rnathuji, line 4: @mjoss17 - Sorry, should have caught this the first time, but let's remove this mapping for course 1 so the test case validates the sparse mapping definitions we expect in practice.

rnathuji (approved):

PR #50, opened 2022-10-17
Add indentation classes

mjoss17:


specs/blocks.spec.ts
```diff
@@ -335,3 +335,27 @@ test('Desmos expressions panel is not visible.', async ({ page }) => {
   await page.waitForSelector('.dcg-grapher')
   await expect(page.locator('text=x=5')).not.toBeVisible()
 })
+
+test('Indentation class indents content correctly', async ({ page }) => {
```
rnathuji, line 339: Good call on including tests for these style classes. Could we add them to a new `styles.spec.ts`, though?

src/styles/indentation.scss
```diff
@@ -0,0 +1,8 @@
+ol.os-raise-noindent {
```
rnathuji, line 1: In terms of organizing these, what do you and @P-Gill97 think about the following options:

1) Creating a `content.scss` with various content style classes
2) Creating sort of a use case specific stylesheet 

The current PRs seem to follow the second approach, but I'm wondering if that's going to result in a lot of small style files which might get a bit confusing after a while :man_shrugging: . If we go with the former we could refactor if / when a single file becomes unwieldy?

mjoss17: I see what you mean, I like the approach of having them together is an encompassing 'style' file and then breaking it apart later as need be.

docs/styles.md
```diff
@@ -0,0 +1,34 @@
...
+
+RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below.
+
+## Style Classes 
+
+### Indent (os-raise-indent)
+Add a padding to the left side of an element. If the element already includes a left padding (like an ordered list) the indent will get set to exactly 32px
+**Example**
+
+**Availiability**
```
rnathuji, line 13: Minor: Spelling issue.

docs/styles.md
```diff
@@ -0,0 +1,34 @@
...
+**Usage**
+```html 
+<p class="os-raise-indent">Text in a box</p>
+```
+
+### No-Indent (os-raise-noindent)
+Remove the indent inherent in an ordered list. 
+**Example**
+
+**Avilibility**
```
rnathuji, line 24: Minor: Spelling issue

docs/styles.md
```diff
@@ -0,0 +1,34 @@
+# Styling content for RAISE
+
+In order to maintain consistency and reliability across all RAISE content, the styling associated with RAISE content is consolidated into a series of css classes located in `src/styles`. 
+
+RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below.
+
+## Style Classes 
+
+### Indent (os-raise-indent)
+Add a padding to the left side of an element. If the element already includes a left padding (like an ordered list) the indent will get set to exactly 32px
```
rnathuji, line 10: This `32px` is actually slightly incorrect. You can refer to [this documentation](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#ems_and_rems) as far as why, but basically if you use `2rem` it will scale with the root font size.

docs/styles.md
```diff
@@ -0,0 +1,34 @@
...
+**Example**
```
rnathuji, line 11: I think you need an empty line before this (it currently shows up on the same line as the preceding sentence).

That being said, did you mean for this to be a parent of `Availability` and `Usage`, or a peer? If the former, maybe it should be `#### Example` :man_shrugging:

docs/styles.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Availiability**
+Any element
+**Usage**
```
rnathuji, line 15: I think you need an empty line before this (it currently shows up on the same line as the preceding sentence).

docs/styles.md
```diff
@@ -0,0 +1,34 @@
...
+```html 
+<p class="os-raise-indent">Text in a box</p>
+```
+
+### No-Indent (os-raise-noindent)
+Remove the indent inherent in an ordered list. 
+**Example**
```
rnathuji, line 22: Same issue / question as previous `Example`.

rnathuji (commented):
My bad -- I didn't look at / review the markdown docs last time, but noticed a few issues on this pass :grimacing: .

docs/styles.md
mjoss17: Ah, I was going to ask @rnathuji  - my thought was that under Example we could add an image of what the final styling product looks like, but I wasn't sure where I should host a screenshot I'd taken and/or where I'd attach it in the repo file structure. 

Is there an easier way to embed images into .md files that are displayed on github that you know of? I was thinking I could k12-openstax content upload them 😂

mjoss17: Alternatively I can just get rid of the Example section for now.

rnathuji: Ah -- screenshots are a nice idea. You can actually just commit them to the repo and reference using a relative path. For example, if you add a file `./docs/static/image1.jpg`, you can then reference it in markdown as `Alt text string`

rnathuji (approved):

PR #51, opened 2022-10-18
Add graybox class for styling

mjoss17:


src/styles/graybox.scss
```diff
@@ -0,0 +1,6 @@
+.os-raise-graybox {
+  border: 1px solid #D9D9D9;
```
rnathuji, line 2: Minor: Up to you as far as whether you want to employ this, but I thought I'd point out that you can define [a variable](https://sass-lang.com/documentation/variables) with the color value. A couple of potential benefits of doing that:

* You can define a human-readable name (e.g. so it's easier to know what #d9d9d9 is)
* You can re-use that same value later

specs/blocks.spec.ts
```diff
@@ -335,3 +335,22 @@ test('Desmos expressions panel is not visible.', async ({ page }) => {
   await page.waitForSelector('.dcg-grapher')
   await expect(page.locator('text=x=5')).not.toBeVisible()
 })
+
+test('Graybox class styles content correctly', async ({ page }) => {
```
rnathuji, line 339: Same as the other PR re: Moving the tests to something like `styles.spec.ts`.

src/styles/graybox.scss
```diff
@@ -0,0 +1,6 @@
+.os-raise-graybox {
```
rnathuji, line 1: Same as the other PR re: potentially moving this to a `content.scss` file.

docs/styles.md
```diff
@@ -0,0 +1,51 @@
...
+
+In order to maintain consistency and reliability across all RAISE content, the styling associated with RAISE content is consolidated into a series of css classes located in `src/styles`. 
+
+RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below. Content Developers 
+
+## Styling Classes 
+
+### Gray Box (os-raise-graybox)
+Add a Gray box around your element
+**Example**
```
rnathuji, line 11: Similar comments on the markdown in this PR as the other as it relates to spelling and missing line breaks:
image

docs/styles.md
```diff
@@ -0,0 +1,73 @@
...
+
+RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below. Content Developers 
+
+
+## Gray Box (os-raise-graybox)
+Add a Gray box around your element
+
+**Example**
+<div style="text-align: center;">
+    <img src="./static/graybox.png" width="350">
```
rnathuji, line 13: I would fix this screenshot per [this comment](https://github.com/openstax/k12/issues/274#issuecomment-1284114779) to not reflect the ability to style interactive blocks as that will break outside of Moodle storage + is not something we want content devs to try and do since the styling / rendering of interactive blocks should be self-contained in the React code.

src/styles/content.scss
```diff
@@ -6,3 +9,10 @@ ol.os-raise-noindent {
 .os-raise-indent {
   padding-left: 2rem;
 }
+
+.os-raise-graybox {
```
rnathuji, line 13: If we want to codify the limitation / expectation that this is for `<div>` we can make this `div.os-raise-graybox` so if it's included on any other element it'll get ignored.

docs/styles.md
```diff
@@ -2,14 +2,42 @@
...
+    <img src="./static/graybox.png" width="450">
+</div>
+
+**Availability**
+Grayboxes should be used in div elements only
+
+
+**Usage** 
+```html 
+<p class="os-raise-graybox">Text in a box</p>
```
rnathuji, line 22: Let's remove this since the class is only allowed in `<div>`

docs/styles.md
```diff
@@ -2,14 +2,39 @@
 
 In order to maintain consistency and reliability across all RAISE content, the styling associated with RAISE content is consolidated into a series of css classes located in `src/styles`. 
 
-RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below.
+RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below. Content Developers
```
rnathuji, line 5: I think you started a sentence that wasn't ever completed here :joy: (`Content Developers...`)

docs/styles.md
```diff
@@ -2,14 +2,39 @@
...
 
-## Style Classes 
 
-### Indent (os-raise-indent)
+## Gray Box (os-raise-graybox)
+Add a Gray box around your element
+
+**Example**
+<div style="text-align: center;">
+    <img src="./static/graybox.png" width="450">
```
rnathuji, line 13: Should we update the screenshot to not include the styled `<p>`? (or I think that's included here at least)

rnathuji (approved):
@mjoss17 - I'm approving this so you can merge once the minor documentation comments are addressed.

PR #52, opened 2022-10-18
Add os-raise-standardtable class for tables

P-Gill97:


src/styles/main.scss
```diff
@@ -1,2 +1,3 @@
 @import "./bootstrap";
+@import "./standardtable";
```
rnathuji, line 2: Let's move this style to the existing `content.scss` file per [this discussion](https://github.com/openstax/k12-apps-raise/pull/50#discussion_r999531259).

src/styles/standardtable.scss
```diff
@@ -0,0 +1,8 @@
+table.os-raise-standardtable {
+  table,
```
rnathuji, line 2: Is this actually necessary? (I think it would apply to a table within a table :man_shrugging: )

rnathuji (approved):

PR #56, opened 2022-11-14
Add EventsAPI Client code

mjoss17:


README.md
```diff
@@ -60,3 +60,20 @@ For `Stylelint`, you will need to add the following to your `.vscode/settings.js
 ## Schemas for Interactive Blocks and Content Templates
 
 Please refer to [schemas](docs/schemas.md) for details on using supported interactive capabilities.
+
+
+## Update the Events API Client Code in EventsAPI
+
+In order to update the client code generated by openapi to match the EventsAPI you can use the commands below. This process creates a docker container with java which runs the openapi-generator-cli commands. 
+
+To update the client code, the command needs to point to the EventsAPI server. However, if the server is being run locally, replace `http://localhost` with `http://host.docker.internal`. To generate client code based on a different openapi source file replace the argument following -i in the command below with the file path or url.
```
rnathuji, line 69: I don't think this approach of using http://host.docker.internal is platform agnostic without some extra magic (it errors on Linux). 

In any case, since we should always be deploying API side changes prior to updating client, let's just document and use `https://events.raiselearning.org/openapi.json`. That way we don't run into any accidental issues from pointing at local API changes.

openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
```
rnathuji, line 1: I think this file is here by mistake. You may have generated it when trying to use the `npm` package, but if I run the docker generator I don't get this file.

README.md
```diff
@@ -60,3 +60,18 @@ For `Stylelint`, you will need to add the following to your `.vscode/settings.js
...
+## Update the Events API Client Code in EventsAPI
+
+In order to update the client code generated by openapi to serve the EventsAPI you can use the command below. This process creates a docker container with java which runs the openapi-generator-cli commands. 
+
+If you want to modify the language or location of the client code, modify the arguments following -g or -o respectively. 
+
+```bash
+$ cd k12-apps-raise
+$ docker run --rm -v "${PWD}:/local" openapitools/openapi-generator-cli generate \
+  -i http://events.raiselearning.org/openapi.json \
```
rnathuji, line 74: Super minor: I mean it'll work / redirect either way, but it would be nice to make this `https` vs `http`.

PR #58, opened 2022-11-29
Moodle api front end client interface

P-Gill97:


src/moodleapi/Moodleapi.ts
```diff
@@ -0,0 +1,39 @@
...
+      args: {
+      }
+    }
+    const url = this.getMoodlePluginMethodUrl(this.endpoint)
+    if (url === '') {
+      console.log('Invalid environment variables')
+    }
+    const response = await fetch(url, { method: 'POST', body: JSON.stringify([userRequest]) })
+    const responseJSON = await response.json()
+
```
rnathuji, line 21: It would be good to do some error handling here. Unfortunately as we found with the service endpoint if there's an error Moodle will not send back an HTTP error code (e.g. `response.ok` will be `True`). Something like the following could be good to include so we don't get a parsing error later.

```javascript
    if (responseJSON.error !== undefined) {
      throw new Error(responseJSON.error)
    }
```

src/moodleapi/Moodleapi.ts
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      uuid: responseJSON[0].data.uuid,
+      jwt: responseJSON[0].data.jwt
+    }
+  }
+
+  getMoodlePluginMethodUrl(endpointMethod: string): string {
+    if (window.M === undefined) {
+      return ''
+    }
```
rnathuji, line 31: Ah, I think this type of logic is why I was thinking earlier we should pass the root and sesskey to the constructor. That way any logic around whether we can even talk to Moodle happens outside by the caller and none of this needs to be here. Instead the below can be:

```javascript
    let userUrl: string
    userUrl = this.wwwroot
    userUrl += '/lib/ajax/service.php?sesskey='
    userUrl += this.sesskey
    userUrl += '&info=' + endpointMethod
    return userUrl
```

Perhaps the constructor then looks like:

```javascript
  private readonly wwwroot: string
  private readonly sesskey: string

  constructor(wwwroot: string, sesskey: string) {
    this.wwwroot = wwwroot
    this.sesskey = sesskey
  }
```

src/moodleapi/Moodleapi.ts
```diff
@@ -0,0 +1,39 @@
...
+    const userRequest = {
+      index: 0,
+      methodname: this.endpoint,
+      args: {
+      }
+    }
+    const url = this.getMoodlePluginMethodUrl(this.endpoint)
+    if (url === '') {
+      console.log('Invalid environment variables')
+    }
```
rnathuji, line 18: This logic can be removed once we incorporate passing the `wwwroot` and `sesskey` in a constructor.

src/moodleapi/Moodleapi.ts
```diff
@@ -0,0 +1,39 @@
+interface Response {
```
rnathuji, line 1: This should be something more specific since we may have other methods later with different response shapes. Maybe the interface name can be `GetUserResponse` :man_shrugging: ?

src/tests/Moodleapi.test.ts
```diff
@@ -0,0 +1,33 @@
...
+afterAll(() => server.close())
+
+test('Test getuser', async () => {
+  const moodleApi = new Moodleapi()
+  window.M = {
+    cfg: {
+      wwwroot: 'http://moodle',
+      sesskey: '12345'
+    }
+  } as any
```
rnathuji, line 28: This can go away with changes to pass values in the constructor.

src/moodleapi/Moodleapi.ts
```diff
@@ -0,0 +1,39 @@
...
+    const response = await fetch(url, { method: 'POST', body: JSON.stringify([userRequest]) })
+    const responseJSON = await response.json()
+
+    return {
+      uuid: responseJSON[0].data.uuid,
+      jwt: responseJSON[0].data.jwt
+    }
+  }
+
+  getMoodlePluginMethodUrl(endpointMethod: string): string {
```
rnathuji, line 28: Minor: What about naming this `getMoodleAjaxUrl` instead? Also, we can mark it `private` since it's an internal helper method.

src/tests/Moodleapi.test.ts
```diff
@@ -0,0 +1,33 @@
+import 'whatwg-fetch'
```
rnathuji, line 1: Similar file rename request here to `MoodleApi.test.ts`.

src/moodleapi/Moodleapi.ts
rnathuji, line 1: Shoving this here for lack of a better place: If we create a `moodleapi/index.ts` with:

```javascript
export * from './Moodleapi'
```

We can then import the class as just:

```javascript
import { MoodleApi } from './moodleapi'
```

:point_up: provides some parity with the import conventions on `eventsapi`, but is a minor subjective nicety.

rnathuji:
@P-Gill97 - Ah, I think you just need to rename `Moodleapi.ts` to `MoodleApi.ts`.

src/moodleapi/Moodleapi.ts
rnathuji: Oh, you can assume it's there. On an error Moodle will include the field (you can test this real quick by changing the method string to something invalid).

src/moodleapi/index.ts
```diff
@@ -0,0 +1,3 @@
+/* tslint:disable */
+/* eslint-disable */
```
rnathuji, line 2: I don't think we actually need these. Removing them will also surface a linter issue for this file not ending with a newline.

src/moodleapi/MoodleApi.ts
```diff
@@ -0,0 +1,45 @@
...
+      throw new Error(responseJSON.error)
+    }
+
+    return {
+      uuid: responseJSON[0].data.uuid,
+      jwt: responseJSON[0].data.jwt
+    }
+  }
+
+  getMoodleAjaxUrl(endpointMethod: string): string {
```
rnathuji, line 37: This probably got lost in the shuffle in my first round of comments, but should we make this a `private` method (e.g. `private getMoodleAjaxUrl(endpointMethod: string): string {`)?

PR #59, opened 2022-12-09
Implement EventManager

mjoss17:


rnathuji (commented):
@mjoss17 , @P-Gill97 - I've marked up various comments in the PR, but per some notes in `events.ts` I think we should pair live this afternoon to discuss some broader changes / feedback I have related to the exported eventing interface and implementation of the `EventManager` class. I think some of the comments I've made here will be impacted by that too, but hopefully any related thrashing will be minimal :crossed_fingers:  .

src/lib/events.ts
```diff
@@ -0,0 +1,139 @@
...
+class EventManager {
+  private static instance: EventManager
+  private static impressionId: string
+  private readonly config: EventManagerConfig | undefined
+  private readonly eventQueue: ApiEvent[] = []
+  private timer: number | undefined
+  private constructor(config?: EventManagerConfig) {
+    this.config = config
+
+    document.addEventListener('visibilitychange', () => {
```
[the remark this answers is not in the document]
mjoss17: I like the idea of limiting the amount of event-dependent calls that are happening on the front end, especially when they're not doing anything. IMO it feels like a cleaner behavior so I'll add this conditional in.

src/lib/events.ts
```diff
@@ -0,0 +1,142 @@
...
+  Configuration,
+  ConfigurationParameters,
+  CreateEventsV1EventsPostRequest
+} from '../eventsapi'
+import { v4 as uuidv4 } from 'uuid'
+import { MoodleApi } from '../moodleapi'
+import { getCurrentContext } from './utils'
+import { ENV } from './env'
+
+export const queueContentLoadedV1Event = async (contentId: string, variant: string): Promise<void> => {
```
rnathuji, line 15: Sorry, I lied, I had one last comment occur to me right when I was about to click this through :sweat_smile: . It's a minor fix, but I realized we have a subtle problem with how / where we generate the timestamp for events. Currently we set that at the very bottom of the call stack. However, due to the async nature of the queuing functions, it's possible for that to make things seem out-of-order relative to how they actually happened. This may not ever actually matter, but in so much as it'd be nice to have event timestamps reflect the closest to the reality of when things happen + to maintain any causal relationships, I think we should change the signatures of the interface to let the callers generate / provide them:

```javascript
export const queueContentLoadedV1Event = async (timestamp: number, contentId: string, variant: string): Promise<void> => {
}

export const queueContentLoadFailedV1Event = async (timestamp: number, contentId: string, error?: string): Promise<void> => {
}
```

...making the usage now:

```javascript
      queueContentLoadedV1Event(Date.now(), contentId, variant).catch((err) => {
        console.error(err)
      })
```

mjoss17: @rnathuji I'm curious if you think it'd be fine to have the date recorded in the interface methods in order to keep the interface as clean as possible for the components. Or do you think getting the timestamp as close as possible to the content load is more important than keeping that interface as simple as possible?

mjoss17: Edit: I see that the way you suggested implementing it still keeps the interface clean (sans timestamp) for the components - I'll plan on this implementation.

rnathuji: I think it'd be good to let callers "own it". We don't have a strong use case right now, but I can imagine scenarios where we may store a timestamp of something that "happens" while collecting some data needed to create that specific event before we can queue it. By poking the timestamp through the interface we allow the callers to make that decision.

rnathuji (approved):

PR #65, opened 2022-12-21
Add styling changes to th and td

P-Gill97:


rnathuji (commented):
@P-Gill97 - This PR looks good to me, but I'm going to hold off on approving so we can wait to discuss the following with Maddy after the holidays:

1. General sanity check that things look the way she wants (with proper markup)
2. Given that I think we'll need to fix tables in our content to use `<th>`, whether she's okay with us merging / deploying this now or wants to have the content edits made first (otherwise we will temporarily have everything top aligned if we deploy the CSS changes first where tables only use `<td>`, but maybe UX is okay with that :man_shrugging: ).

rnathuji:
@madisonschear - @P-Gill97 can pair / demo the styling with you to be sure it all looks good before we merge, but we also had a logistical question (see item 2 [here](https://github.com/openstax/k12-apps-raise/pull/65#pullrequestreview-1226722548)) we wanted your feedback on.

mjoss17 (approved):

rnathuji (approved):

PR #68, opened 2023-01-10
Improve event setup behavior for multiple async loads

rnathuji:


src/lib/events.ts
```diff
@@ -60,7 +62,17 @@ class EventManager {
...
-    const user = await moodleApi.getUser()
+
+    // Wait on a new or existing promise for user data from Moodle
+    this.getUserPromise = this.getUserPromise ?? moodleApi.getUser()
+    const user = await this.getUserPromise
+
+    // Check for an initialized instance again in case there were multiple
+    // waiters on the promise
+    if (this.instance !== undefined) {
+      return this.instance
```
rnathuji, line 73: Just as a note during review: I'm not sure why coverage isn't picking this up. If you comment this line the unit test will fail, and if I drop a break point here it will hit while running the test :man_shrugging: .

mjoss17 (approved):

src/lib/events.ts
rnathuji: I ended up figuring out that this has to do with an interaction between the singleton setup that occurs in `events.ts` and ordering of tests. The "fix" I've incorporated for now is just to move the new test to the top so it gets run first. This is definitely a hack and not something I'm proud of, but a better alternative didn't immediately occur to me and I figure it's better than no test case / coverage :man_shrugging: .

PR #73, opened 2023-02-06
Add parsing for contentid

mjoss17:


src/lib/blocks.tsx
```diff
@@ -172,6 +172,7 @@ export const parseUserInputBlock = (element: HTMLElement): JSX.Element | null =>
   const maybeButtonText = element.dataset.buttonText
   const waitForEvent = namespaceEvent(element.dataset.waitForEvent)
   const fireEvent = namespaceEvent(element.dataset.fireEvent)
+  const dataContentId = element.dataset.dataContentId
```
rnathuji, line 175: Sorry for not catching this during the screen share yesterday, but all of these instances in the PR should be renamed to just `contentId` like how we parse this attribute [here](https://github.com/openstax/k12-apps-raise/blob/cf746ce58953ab86c9102c5b16999fda61f4b884/src/lib/content.tsx#L13). That's because the `data-` gets stripped automatically (e.g. `data-fire-event` becomes `fireEvent`, `data-content-id` becomes `contentId`, etc.). You can find documentation on this behavior [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) if you're curious.

In good news, I think you can fix this via find / replace in vscode pretty easily :sweat_smile: .

rnathuji (approved):

PR #83, opened 2023-05-02
Implement table styling for v2

MReyna12:


MReyna12:
As of May 8, 2023 at 3:40 PM CST I am only awaiting confirmation from Maddy that the skinny table width of 240px is okay. The aforementioned item is the only thing outstanding outside of any suggested changes by the reviewer.

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+  }
+
+  th,
+  td {
+    text-align: center;
+    padding: 16px;
+  }
+
+  th {
+    font-size: 1.2rem;
```
rnathuji, line 91: This may need to be `1.125rem` if we're targeting `18px` (currently it is `19.2px`).

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+    font-weight: 700;
+    color: #fff;
+  }
+
+  td {
+    font-size: 1.067rem;
```
rnathuji, line 97: I think this should be `1rem`.

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+  }
+
+  thead tr th:first-child {
+    border: none;
+    background-color: #fff;
+  }
+
+  td {
+    color: $raise-light-black;
+    font-size: 1.067rem;
```
rnathuji, line 137: This should perhaps be `1rem` for `16px`.

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+
+  tr:first-child td {
+    border-top: none;
+  }
+}
+
 .os-raise-d-flex {
   display: flex;
   flex-wrap: wrap;
+  row-gap: 5rem;
```
rnathuji, line 233: I'm curious as to the value of `5rem` here since that'll convert to `80px` which ends up looking like:

image

This might be worth a quick check with UX to see if it looks okay :man_shrugging: .

src/styles/content.scss
```diff
@@ -1,6 +1,9 @@
 $raise-graybox: #f5f5f5;
```
rnathuji, line 1: General comment on this PR which I'm putting here for no particular reason: I noticed there are lots of small commits in the history. This isn't necessarily a bad thing, and it's definitely a good habit to commit often when you're developing. However, sometimes it's nice to keep the history we push / merge a bit cleaner by creating a more intentional set of commits. One workflow to accomplish that is to perform an interactive rebase to squash / consolidate commits as well as re-order as necessary. If this is something you wan to potentially look at and the `rebase` workflow is new for you, we can pair to do that once this PR is otherwise approved (one of our conventions is that devs refrain from rebasing during a review since it impacts local branches that may have been pulled).

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+    text-align: left;
+  }
+
+  td:first-child {
+    border-left: none;
+  }
+
+  p {
+    margin-bottom: 0;
+  }
```
rnathuji, line 160: I'm not sure if this is a good idea as it will break something like:

```html
      <td>
        <p>A11(B) simplfy numeric and algebraic expressions using the laws of exponents, including integral and rational exponents</p>
        <p>A11(B) simplfy numeric and algebraic expressions using the laws of exponents, including integral and rational exponents</p>
      </td>
```

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+  max-width: 240px;
+
+  td:first-child {
+    border-left: none;
+  }
+}
+
+.os-raise-skinnytablemedia {
+  margin: auto;
+}
```
rnathuji, line 191: Was this class perhaps from testing / tinkering?  I see it's also used in the example HTML, but I wanted to confirm whether it was intentionally included. If so, maybe we should consider a more general utility class for allowing authors to apply `margin: auto` when desired :man_shrugging: ? If we go that route we should document the new utility class explicitly as well.

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+  }
+}
+
+table.os-raise-doubleheadertable {
+  table-layout: fixed;
+  width: 100%;
+
+  caption {
+    color: $raise-light-black;
+    font-size: 1.6rem;
```
rnathuji, line 111: Same comment as above as it relates to possibly correcting `1.6rem` to be `1.5rem`.

rnathuji (commented):
Things look really nice in general, @MReyna12 ! :tada:  I have a few minor comments here and there on the code, as well as a general comment on potentially cleaning up some of the commit history.

src/styles/content.scss
```diff
@@ -57,9 +60,177 @@ table.os-raise-standardtable {
...
+     radius applied to the table)
+  */
+  border: 2px solid $raise-table-color;
+  border-collapse: separate;
+  border-spacing: 0;
+  border-radius: 5px;
+
+  caption {
+    color: $raise-light-black;
+    font-size: 1.6rem;
```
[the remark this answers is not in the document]
MReyna12: I will address the overall px to rem font changes in this comment.

The reason I used a base of 15px is because the body font-size is .9375rem, so I interpreted that as meaning the base is 15px. I also tried to compare text using the px/rem values next to each other and it seemed like the current rem values matched the former px values, but that test was harder to make a determination because the difference is so small. Happy to change the rem values as you state throughout your comments, but wanted to note the above prior to making the change.

src/styles/content.scss
MReyna12: I have no preference in regards to how much space there is between rows, so I just put a value in that I thought created enough space between the tables.

@madisonschear - Whenever the browser window is not large enough to allow the tables to be on the same row they will drop down a row. Previously we didn't have any space between the tables when they dropped down a row so I incorporated that concept here. Could you let me know if 5rem works for you or if you'd like a different value?

src/styles/content.scss
MReyna12: When Prabhdip was helping me out on an unrelated issue last week he also mentioned the commit conventions in the engineering docs. I've since made my commits align with the doc and intended on cleaning up the commit history after the PR is approved. I have only used rebase maybe a handful of times, so I think it would be very helpful to pair whenever the PR is approved so I can make sure I rebase correctly.

src/styles/content.scss
MReyna12: I was not sure if it made more sense to make a general utility class or just keep it as a "companion" class. I am okay with making a general utility class. Since we are adjusting the margin to center the skinny tables, I also needed to be able to center the image/graph or else when the table wraps under the image/graph only the table was being centered.

rnathuji: I'd lean towards defining / using a utility class in the example. It might come in handy for authors in other circumstances (e.g. to center a stand-alone image or something 🤷‍♂️ ). If you wanted to steal bootstrap naming conventions, it could be `os-raise-mx-auto`, but open to other suggestions / ideas.

src/styles/content.scss
MReyna12: Agreed - the reasoning behind this was to try and align the design to what Maddy provided. I can revert back. Below is the difference in spacing w/ and w/o the margin-bottom.

Without margin-bottom:
image


With margin-bottom (default 1rem):
image

rnathuji: :thinking: -- I suppose in this specific type of table one could do something like the following to avoid the margin added by `<p>`:

```html
<table class="os-raise-textheavytable">
  <caption> Texas Essential Knowledge and Skills (TEKS)
  <thead>
    <tr>
      <th scope="col">TEKS</th>
      <th scope="col">Explination of Coverage</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        A11(B) simplfy numeric and algebraic expressions using the laws of exponents, including integral and rational exponents
      </td>
      <td>
        Partial coverage: Lesson provides content that covers part of this TEKS. The parts that are covered have been underlined
      </td>
    </tr>
  </tbody>
</table>
```

But, that doesn't work so well when you do want `<p>`s in the cell. Perhaps the general case warrants creating a utility class like `os-raise-mb-0` :man_shrugging: ?

src/styles/content.scss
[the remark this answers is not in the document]
MReyna12: That is extremely helpful! I'll keep that in mind for the future and go ahead and adjust the rem values.

src/styles/content.scss
rnathuji: :+1: -- Yeah, that's what I'm thinking.

src/styles/content.scss
madisonschear: This works for me.

madisonschear: Can we set the spacing to shrink based on the browser window? So that it shrinks to a minimum of spacing before dropping down to a new row?

MReyna12: Are you referring to the spacing as seen in the screenshot? If so, then that spacing will already shrink until the tables are just about touching and then a table drops to a new row.

verticalspacing

docs/styles.md
```diff
@@ -2,80 +2,507 @@
 
 In order to maintain consistency and reliability across all RAISE content, the styling associated with RAISE content is consolidated into a series of css classes located in `src/styles`. RAISE Style classes are either written for specific elements, or can be applied to a variety of objects. Those differences are listed below.
```
rnathuji, line 3: A couple of comments related to doc updates:

1. Now that this doc is starting to get kind of long, what do you think about adding a ToC like the following to the top of the doc (you may want to use a different structure / labels -- this is just to illustrate the suggestion):

```markdown
Style classes:

* Tables
  * [.os-raise-textheavytable](#text-heavy-table)
  * [.os-raise-horizontaltable](#horizontal-table)
  * [.os-raise-midsizetable](#mid-size-table)
  * `.os-raise-skinnytable`
    * [Example: .os-raise-skinnytable with media](#skinny-table-table--media)
    * [Example: .os-raise-skinnytable multiple tables](#skinny-table-three-tables)
  * [.os-raise-wideequaltable](#wide-equal-column-table)
  * [.os-raise-wideadjustedtable](#wide-adjusted-table)
  * [.os-raise-doubleheadertable](#double-header-table)
* Spacing
  * [.os-raise-indent](#indent-os-raise-indent)
  * [.os-raise-noindent](#no-indent-os-raise-noindent)
  * [.os-raise-text-center](#centered-text-os-raise-text-center)
* Layout
  * [.os-raise-d-flex](#side-by-side-os-raise-d-flex)
  * [.os-raise-align-items-center](#side-by-side-os-raise-d-flex)
  * [.os-raise-justify-content-center](#side-by-side-os-raise-d-flex)
  * [.os-raise-justify-content-between](#side-by-side-os-raise-d-flex)
  * [.os-raise-justify-content-evenly](#side-by-side-os-raise-d-flex)
* Components
  * [.os-raise-graybox](#gray-box-os-raise-graybox)
  * [.os-raise-motivation](#motivational-content-os-raise-motivation)
```

2) Let's add documentation for the two new utility classes you have including `.os-raise-mb-0` and `.os-raise-mx-auto`

rnathuji (commented):
@MReyna12 - This looks good to me aside from a couple of small doc comments. 

Let's pair tomorrow to work on the commit restructuring. Also, if you remind me, as part of that we can add in a commit to update the CI action versions too.

docs/styles.md
MReyna12: My thoughts on your points below:

1. I think a ToC makes sense. I will incorporate one into the markdown.

2. I realized I forgot to include those in the documentation after I pushed the commit. Just finished my suggested documentation adds for those two utility classes.

docs/styles.md
```diff
@@ -1,81 +1,594 @@
...
+  <thead>
+    <tr>
+      <th></th>
+      <th scope="col">\(x\)</th>
+      <th scope="col">\(+7\)</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th scope="rowl">\(x\)</th>
```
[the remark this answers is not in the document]
MReyna12: Thanks for pointing this out. I've fixed the typo.

docs/styles.md
```diff
@@ -1,81 +1,594 @@
...
+
+Add as a class attribute to a table html tag.
+
+**Usage**
+
+```html
+<table class="os-raise-doubleheadertable">
+  <thead>
+    <tr>
+      <th></th>
```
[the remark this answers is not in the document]
MReyna12: Thanks for pointing this out. I've added the scope attribute.

rnathuji (approved):
Approving from my end.

PR #89, opened 2023-05-25
V2 ib style updates

P-Gill97:


src/styles/bootstrap.scss
```diff
@@ -105,4 +112,57 @@
...
+  }
+
+  .os-btn {
+    /* stylelint-disable-next-line scss/at-extend-no-missing-placeholder */
+    @extend .btn;
+
+    padding: 8px 32px;
+  }
+
+  .os-raise-bootstrap label {
```
rnathuji, line 163: Given where this is, can't it just be?:

```css
  label {
    display: flex;
    align-items: center;
  }
```

It seems an implication of :point_up: is any label that is ever in `.os-raise-bootstrap` will get this styling. Should we be more selective to future proof ourselves a bit? For example, `CustomCheckbox` could support a prop that allows specifying classes to apply to the label. With that approach we can actually avoid having this here altogether and define a class in `interactives.scss`. :man_shrugging:

src/styles/interactives.scss
```diff
@@ -0,0 +1,96 @@
...
+
+p.os-raise-attempts-text {
+  color: $raise-medium-gray;
+}
+
+div.os-raise-justify-content-end {
+  justify-content: end;
+}
+
+.disabled {
```
rnathuji, line 81: This is just an example, but I feel like we have some inconsistency in our class naming conventions currently. Here is what (I think) we were doing before:

* Classes that are "public" (e.g. documented for authors and defined in `content.scss`): Prefix with `os-raise-`
* Classes that were "internal" (previously there were just [these two](https://github.com/openstax/k12-apps-raise/blob/4c9ca86e9adab9272288c6b90d98582d0e819d5b/src/styles/bootstrap.scss#L95-L107), so it's arguable whether that constitutes a "convention" or just inconsistency): Prefix with `os-`
    * I guess `.os-raise-bootstrap` kind of violates this "convention" :thinking: 

In any case, thoughts on what we want to try and do moving forward @P-Gill97 , @MReyna12 ? Maybe everything should be `os-raise-*`? Or do we want to differentiate "internal" vs "external" somehow?

src/styles/bootstrap.scss
MReyna12: I can't recall the exact reasoning that I wrote the above CSS, but it is not necessary to include in the `bootstrap.scss` file. Thanks for catching this.

Instead, I can add `display: flex` and `align-items: center` to `label.os-raise-fill-label-container` (maybe change the class name to `label.os-raise-label-container`) in `interactives.scss`. I think would resolve our issue and wouldn't require use of a prop in CustomCheckbox.

src/styles/interactives.scss
MReyna12: Regarding this specific example - I can revise to `os.raise-disabled`. 

I believe you and/or Prabhdip may have mentioned the public/internal concept, but I may have misunderstood the prefix `os-raise` versus `os-`. 

So public classes are limited to classes in `content.scss`. This would not include styles in `interactives.scss`, correct? Since these classes would be only for engineer use and not content authors. If this is the case and we want to differentiate between internal / external, then I am okay with the `os-` prefix for internal classes and `os-raise-`.

MReyna12:
@rnathuji @P-Gill97 - I've gone through and updated the class names and consolidated various classes. I was also able to solve the specificity issue I previously mentioned - this allowed me to create a more organized `interactives.scss` style sheet. The `bootstrap.scss` style sheet now contains only classes that explicitly `@extend` a bootstrap class. I also updated the applicable components and tests as it related to the CSS changes.

Happy to discuss here and/or during our pairing session tomorrow.

docs/styles.md
```diff
@@ -13,6 +13,7 @@
     - [Equal Column Wide Tables](#wide-equal-column-table)
```
rnathuji, line 13: It looks like somehow, in the confusion of the various branches :sweat_smile: , we accidentally backed out the changes in [this PR](https://github.com/openstax/k12-apps-raise/pull/91/files) (e.g. if we were to merge this PR we'd revert the changes there). The specific commits that seem to impact things:

* [b9eff6aa](https://github.com/openstax/k12-apps-raise/pull/89/commits/b9eff6aa0aecfe199163ac66a457542a14552378#diff-241f5064b93adeb15ef8ad6eb6db4f042d686554aef78819feea29a98a17db5d): This reverted the changes in `content.scss`
* [deb150b96](https://github.com/openstax/k12-apps-raise/pull/89/commits/deb150b96d5fe9729888f72ff450344cbb636583): This reverted the corresponding doc changes.

Since the commits :point_up: include other changes, just reverting them on this branch is likely a pain. One way to fix it could be to just cherry pick the commit from PR 91 to this branch, though that would impact history a bit:

```
$ git cherry-pick f3e8cc0e7369a32b36c9cff2b8e30982d697603b
```

In any case, something we'll need to address somehow prior to merging.

src/styles/bootstrap.scss
```diff
@@ -68,9 +68,16 @@
...
+  --bs-font-sans-serif: system-ui, -apple-system, "Segoe UI", roboto,
+    "Helvetica Neue", "Noto Sans", "Liberation Sans", arial, sans-serif,
+    "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  --bs-font-monospace: sfmono-regular, menlo, monaco, consolas,
+    "Liberation Mono", "Courier New", monospace;
+  --bs-gradient: linear-gradient(
+    180deg,
+    rgb(255 255 255 / 15%),
+    rgb(255 255 255 / 0%)
+  );
```
rnathuji, line 80: I just want to confirm my :eyes: here: We didn't make any changes to these lines, it's just some formatting change, correct? I ask because these are copied from Bootstrap files and can get overwritten when we update versions of Bootstrap. I want to make sure doing that in the future would break anything.

MReyna12: This is just a formatting change - I believe it occurred prior to me removing prettier.

MReyna12:
@madisonschear - Please see below for examples of what the icons look next to the dropdown and input interactive. Let us know if these are good as-is or if you would like any additional updates.

**Dropdown**

image

image


**Input**

image

image

src/components/AttemptsCounter.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+interface Props {
+  retryLimit: number
+  retriesAllowed: number
+}
+
+export const AttemptsCounter: React.FC<Props> = ({ retryLimit, retriesAllowed }) => (
```
rnathuji, line 8: Super minor: This declaration is somewhat inconsistent in this code base.

src/components/Icons.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+
+interface IconProps {
+  className?: string
+}
+
+export const CorrectAnswerIcon: React.FC<IconProps> = ({ className }) => (
```
rnathuji, line 7: Super minor: This declaration is somewhat inconsistent in this code base.

src/components/Icons.tsx
```diff
@@ -0,0 +1,35 @@
...
+    className={className}
+  >
+    <path
+      fill="#019920"
+      d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10s10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5l1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z"
+    />
+  </svg>
+)
+
+export const WrongAnswerIcon: React.FC<IconProps> = ({ className }) => (
```
rnathuji, line 22: Super minor: This declaration is somewhat inconsistent in this code base.

src/components/AttemptsCounter.tsx
```diff
@@ -0,0 +1,23 @@
...
+                </p>
+            )
+          : (
+                <p className="os-attempts-text">
+                    Attempts left: {retryLimit - retriesAllowed + 1}/
+                    {retryLimit + 1}
+                </p>
+            )}
+    </div>
+)
```
rnathuji, line 23: Minor / white space inconsistency: The indentation here is 4 spaces vs 2.

src/components/CustomCheckbox.tsx
```diff
@@ -0,0 +1,55 @@
...
+  label: string
+  type: string
+  correct: boolean
+  disabled: boolean
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
+  showAnswer: boolean
+  selected: boolean
+}
+
+export const Checkbox = ({
```
rnathuji, line 13: Minor: I think we discussed the issue / challenge of coming up with a name for this that maintains clarity early on. My only comment with the current version is it is somewhat inconsistent relative to our other files where the component and filename match (the new `Icons.tsx` is an exception, but that makes natural sense due to it including multiple components).

Thoughts on a potential alternative name in case y'all want to rename it altogether:
* `Selectable`
* `SelectableOption`
* `SelectableChoice`
* `SelectableBox`
* `SelectableAnswer`
* `FieldBox`
* `FormSelectable`
* `AnswerBox`

rnathuji (approved):
@MReyna12 , @P-Gill97 - I took a pass through and only had a few minor / subjective comments related to consistency, white space, and one potential naming nit. I'll defer to y'all as far as if / how any of those are worth addressing prior to merging and am going to approve from my side.

src/components/AttemptsCounter.tsx
MReyna12: Fixed.

image

src/components/CustomCheckbox.tsx
MReyna12: I like `SelectableAnswer` or `FormSelectable` - leaning more towards `FormSelectable`.

src/components/AttemptsCounter.tsx
MReyna12: This one is probably extremely obvious, but it is going over my head. What is inconsistent about the declaration?

rnathuji: Ah -- just the use of `React.FC<Props>` to define the type. It might be "more" consistent as:

```javascript
export const AttemptsCounter = ({ retryLimit, retriesAllowed }: Props): JSX.Element => (
```

MReyna12: That makes sense. I'll revise accordingly.

MReyna12: Updated per this [commit](https://github.com/openstax/k12-apps-raise/pull/89/commits/040f8931d9c2d6451b0d23a1920bccde6a4843cd).

src/components/Icons.tsx
MReyna12: Updated per this [commit](https://github.com/openstax/k12-apps-raise/pull/89/commits/040f8931d9c2d6451b0d23a1920bccde6a4843cd).

src/components/Icons.tsx
MReyna12: Updated per this [commit](https://github.com/openstax/k12-apps-raise/pull/89/commits/040f8931d9c2d6451b0d23a1920bccde6a4843cd).

src/components/AttemptsCounter.tsx
MReyna12: Updated per this [commit](https://github.com/openstax/k12-apps-raise/pull/89/commits/040f8931d9c2d6451b0d23a1920bccde6a4843cd).

src/components/CustomCheckbox.tsx
MReyna12: Updated per this [commit](https://github.com/openstax/k12-apps-raise/pull/89/commits/040f8931d9c2d6451b0d23a1920bccde6a4843cd).

PR #92, opened 2023-06-01
Update styles for input problem

P-Gill97:


src/styles/bootstrap.scss
```diff
@@ -142,7 +142,30 @@
...
+  }
+
+
+
+  .os-form-control-correct-answer-choice {
+    border-color: #019920;
+  }
+
+  .os-form-control:disabled {
+    box-shadow: none;
```
rnathuji, line 168: A couple of general comments on these styles:

1. We may not want to hard-code some of these constant values and instead use / define variables like we do [here](https://github.com/openstax/k12-apps-raise/blob/ab10977c533cf119e74c3caf207b084d9c1c8b82/src/styles/interactives.scss#L1-L8)
2. I feel like there may be some redundancy with some of these classes being added in `bootstrap.scss` and some of the ones we're creating in `interactives.css` such as [here](https://github.com/openstax/k12-apps-raise/blob/ab10977c533cf119e74c3caf207b084d9c1c8b82/src/styles/interactives.scss#L34-L37). Perhaps this is a known thing that you and @MReyna12 have already discussed and plan to consolidate once everything is in the same branch, but just wanted to point it out / ask :man_shrugging: .

MReyna12: Going to both points - there is likely a number of areas of redundancy. The intent was to comb through redundant CSS post-merge into the v2-ib-style-updates branch and consolidate then. There may be some redundancy in `bootstrap.scss` and `interactives.scss` due to specificity issues caused by the os-raise-bootstrap class.

rnathuji (approved):

PR #95, opened 2023-06-15
Add vertical alignment and updated png

MReyna12:


MReyna12:
I noticed the screenshot of the `textheavytable` was an older version that had the word "Explanation" incorrectly spelled (Explination) so I also updated that screenshot as part of this PR.

src/styles/content.scss
```diff
@@ -151,6 +151,10 @@ table.os-raise-textheavytable {
     text-align: left;
   }
 
+  td {
```
rnathuji, line 154: @MReyna12 - This was a bit hidden in the issue content, but I think we need the following as well to address "The headers of the textheavytables is still aligned to the bottom vertically":

```css
  th {
    vertical-align: bottom;
  }
```

MReyna12: Added via this [commit](https://github.com/openstax/k12-apps-raise/pull/95/commits/12befc628129808d9a5f9cc824c67d75decbfe9b).

rnathuji (approved):

PR #97, opened 2023-06-21
Add focus border color

MReyna12:


MReyna12:
As part of this PR I also removed the hover styling once the question has been submitted correctly / attempts are exhausted. I added it here because it was part of the original design implementation (like the focus border), but once the class name builder function was introduced we forgot to add the class that disables hover effects once a question is submitted correctly / attempts are exhausted.

rnathuji (approved):

PR #99, opened 2023-06-26
Update dependency versions

rnathuji:


src/components/MultipleChoiceProblem.tsx
```diff
@@ -79,13 +79,14 @@ export const MultipleChoiceProblem = ({
   const generateOptions = (
     values: MultipleChoiceFormValues,
     isSubmitting: boolean,
-    setFieldValue: (field: string, value: any, shouldValidate?: boolean) => void
+    // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+    setFieldValue: (field: string, value: any, shouldValidate?: boolean) => Promise<void | FormikErrors<MultipleChoiceFormValues>>
```
MReyna12, line 83: Are you disabling this es-lint rule because typically void cannot be mixed with any other types as a return value?

MReyna12 (approved):

PR #101, opened 2023-07-14
Mathfield component

P-Gill97:


specs/blocks.spec.ts
```diff
@@ -305,6 +305,39 @@ test('attempts-exhausted response overrides answer-specific response', async ({
...
+  </div>
+  `
+  await mockPageContentRequest(page, htmlContent)
+  await page.goto('/')
+  await page.locator('.ML__content').click()
+  await page.waitForSelector('text=y')
+  await page.keyboard.type('42')
+  await page.waitForSelector('text=42')
+  await page.locator('text=Check').click()
+  await page.locator('text=Check').click()
```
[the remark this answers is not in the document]
MReyna12: Per this [comment](https://github.com/openstax/k12/issues/434#issuecomment-1646270944) I have managed to fix the need for users to click the check button more than once to submit an answer. I have removed the second `await page.locator('text=Check').click()` and the test passes as expected.

specs/blocks.spec.ts
```diff
@@ -305,6 +305,38 @@ test('attempts-exhausted response overrides answer-specific response', async ({
...
+      <p id="correct">Correct response with math: \\( x=2 \\)</p>
+    </div>
+    <div class="os-raise-ib-pset-encourage-response">
+      <p id="encourage">Encourage response with math: \\( x=2 \\)</p>
+    </div>
+  </div>
+  `
+  await mockPageContentRequest(page, htmlContent)
+  await page.goto('/')
+  await page.locator('.ML__content').click()
```
rnathuji, line 332: I think we can just make this selector `'math-field'`. That would probably be less brittle + test for the expected element itself.

specs/blocks.spec.ts
```diff
@@ -305,6 +305,38 @@ test('attempts-exhausted response overrides answer-specific response', async ({
...
+  await page.waitForSelector('text=y')
```
rnathuji, line 333: It took me a second, but I think the intent of this is to basically assert that the keyboard appears. If so, perhaps using `'.ML__keeyboard'` as the selector would make that a bit clearer + not subject to failure if we were to remove `y` from the keyboard :man_shrugging: ?

src/components/Mathfield.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface MathfieldProps {
+  onInput: (event: React.ChangeEvent<MathfieldElement>) => void
+  className: string
+  disabled: boolean
+}
+
+export const Mathfield = ({ className, disabled, onInput }: MathfieldProps): JSX.Element => {
+  const mathfieldRef = useRef<MathfieldElement>(null)
+  useEffect(() => {
```
rnathuji, line 24: It's a minor thing, but we may want to split this up into two `useEffect` hooks where we use an empty dependency array for the things which we don't want to run every time `disabled` might change:

```javascript
  useEffect(() => {
    const mathFieldCurrent = mathfieldRef.current
    if ((mathFieldCurrent === null)) {
      return
    }
    mathFieldCurrent.mathVirtualKeyboardPolicy = 'manual'

    mathFieldCurrent.addEventListener('focusin', (ev) => {
      window.mathVirtualKeyboard.show()
    })

    mathFieldCurrent.addEventListener('focusout', (ev) => {
      window.mathVirtualKeyboard.hide()
    })
  }, [])

  useEffect(() => {
    const mathFieldCurrent = mathfieldRef.current
    if ((mathFieldCurrent === null)) {
      return
    }

    if (disabled && mathFieldCurrent.shadowRoot !== null) {
      const pointerEventSpan = mathFieldCurrent.shadowRoot.querySelector('span')
      if (pointerEventSpan !== null && pointerEventSpan !== undefined) {
        pointerEventSpan.style.pointerEvents = 'none'
      }
    }
  }, [disabled])
```

src/components/InputProblem.tsx
```diff
@@ -132,12 +134,26 @@ export const InputProblem = ({
...
-                disabled={inputDisabled || isSubmitting}
-                autoComplete={'off'}
-                onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => { clearFeedback(); void setFieldValue('response', e.target.value) }}
-                className={buildClassName(values.response, solution, inputDisabled || isSubmitting)} />
+              {
+                comparator.toLowerCase() === 'math'
+                  ? (
+                  <Field
+                    name="response"
+                    disabled={inputDisabled}
```
rnathuji, line 142: I know I discussed this with @MReyna12 , but what goes wrong if we make this `disabled={inputDisabled || isSubmitting}` again?

src/components/InputProblem.tsx
```diff
@@ -132,12 +134,26 @@ export const InputProblem = ({
...
+                    as={Mathfield}
+                    onInput={(e: React.ChangeEvent<MathfieldElement>): void => { clearFeedback(); void setFieldValue('response', e.target.value); console.log('This is the value:', e.target.value) }}
```
rnathuji, line 144: I think this has a debug `console.log` which we probably want to remove.

src/styles/bootstrap.scss
```diff
@@ -167,7 +167,7 @@
   .os-form-control {
     /* stylelint-disable-next-line scss/at-extend-no-missing-placeholder */
     @extend .form-control;
-
+
```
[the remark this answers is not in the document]
MReyna12: I was getting a linter error for not having that extra space so I added the extra space. If you notice the rest of the classes in `bootstrap.scss` also have a space right after the `@extend` line.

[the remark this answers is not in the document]
rnathuji: I think you get a linter error if the empty line isn't there, but there are a few spaces on the line just hanging out. I typically set the following in my VSCode settings to see / auto-trim these:

```
"editor.renderWhitespace": "all",
"files.trimTrailingWhitespace": true,
```

src/components/Mathfield.tsx
MReyna12: That's a good catch, you are absolutely right.

src/components/InputProblem.tsx
rnathuji: But there's no buggy behavior if we do that? I ask because using `inputDisabled || isSubmitting` is the more technically correct approach. For example, if you add `await new Promise(r => setTimeout(r, 10000))` to `handleSubmit()`, currently while the handler is "running" a user can continue interacting with the input which is incorrect. I vaguely recalled us discussing not including `isSubmitting` due to some issue that occurred when we did, and the "race condition" issue created being the lesser issue. However, if we don't currently see problems / have a specific reason not to, I think we should keep parity with the other code path here and use `disabled={inputDisabled || isSubmitting}` as that should address both scenarios (form handler running and user attempts exhausted).

specs/blocks.spec.ts
MReyna12: Is this what you are ultimately suggesting between this comment and [this](comment)?

```js
await mockPageContentRequest(page, htmlContent)
  await page.goto('/')
  await page.locator('math-field').click()
  await page.waitForSelector('.ML__keyboard')
  await page.keyboard.type('42')
  await page.waitForSelector('text=42')
  await page.locator('text=Check').click()
  await page.waitForSelector('text=Correct')
```

MReyna12: I think the reason Prabhdip did not use `math-field` as the locator is because when you click the `math-field` what you are actually clicking is a `<span>` inside of `math-field` with the class `.ML__content`. I tried making `math-field` the locator but the test failed when it tried to click the `math-field`.

The following test does pass:
```js

Comment (rnathuji) [reply specs/blocks.spec.ts:333]: :confused: -- I seem to get the following passing:

```javascript
  await mockPageContentRequest(page, htmlContent)
  await page.goto('/')
  await page.locator('math-field').click()
  await page.waitForSelector('.ML__keyboard')
  await page.keyboard.type('42')
  await page.waitForSelector('text=42')
  await page.locator('text=Check').click()
  await page.waitForSelector('text=Correct')
```

The output screenshot seems correct so it doesn't look like a false positive pass :thinking:  : 

image

src/styles/bootstrap.scss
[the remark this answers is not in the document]
MReyna12: Ahhhh I see what you are referring to now. I've removed those spaces but kept the new line.

I'll make those changes in my VSCode too. Admittedly, I am still adjusting to not having Prettier handle all the space clean-ups for me as you can tell throughout this PR 😂

src/components/InputProblem.tsx
rnathuji: Discussed live but noting here: The buggy behavior I was overlooking was that with `inputDisabled || isSubmitting`, after the first attempt you couldn't modify things anymore. After discussion we decided to just make a minor change to the conditional logic in the `useEffect()` hook for `disabled` to address that.

specs/blocks.spec.ts
rnathuji: Discussed live but noting here: It turns out the test was passing sometimes on @MReyna12 's machine, and the failures were due to it hitting a timeout. We're going to try the `'math-field'` selector and see if things pass on CI.

src/components/InputProblem.tsx
MReyna12: Per a slack call with @rnathuji, we've opted to add the following code to allow for `isSubmitting` to be added back as a value of `disabled` within `InputProblem.tsx`.

```js
if (!disabled && mathFieldCurrent.shadowRoot !== null) {
      const pointerEventSpan = mathFieldCurrent.shadowRoot.querySelector('span')
      if (pointerEventSpan !== null && pointerEventSpan !== undefined) {
        pointerEventSpan.style.pointerEvents = 'auto'
      }
    }
```

specs/blocks.spec.ts
MReyna12: Per a slack call with @rnathuji, we identified that while the test operates correctly, sometimes a timeout error is reached because it takes longer than 10000ms to complete the test. Per the documentation on `locator().click()` (found [here](https://playwright.dev/docs/api/class-locator#locator-click)), when all steps combined have not finished during the specified timeout (aka 10000ms), this method throws a [TimeoutError](https://playwright.dev/docs/api/class-timeouterror).

src/components/InputProblem.tsx
MReyna12: Fixed in latest PR.

src/global.d.ts
```diff
@@ -13,6 +16,12 @@ declare global {
       }
     }
   }
+  namespace JSX {
+    interface IntrinsicElements {
+      ['math-field']: Partial<MathfieldElementAttributes & DOMAttributes<MathfieldElement>>
+    }
+  }
+
```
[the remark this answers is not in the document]
MReyna12: Fixed in latest PR.

jest.config.ts
```diff
@@ -8,5 +8,10 @@ export default {
     '!src/*.d.ts',
     '!src/lib/env.ts'
   ],
+  moduleNameMapper: {
+    '^mathlive$': '<rootDir>/node_modules/mathlive'
+  },
+  setupFiles: ['<rootDir>/jest.setup.ts'],
   verbose: true
+
```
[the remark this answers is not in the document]
MReyna12: Fixed in latest PR.

jest.setup.ts
```diff
@@ -0,0 +1,6 @@
+global.ResizeObserver = jest.fn().mockImplementation(() => ({
+    observe: jest.fn(),
+    unobserve: jest.fn(),
+    disconnect: jest.fn(),
+  }))
+
```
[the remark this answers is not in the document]
MReyna12: Fixed in latest PR.

rnathuji (approved):

PR #102, opened 2023-07-26
Add support for math input problem 

P-Gill97:


src/components/InputProblem.tsx
```diff
@@ -67,14 +68,35 @@ export const InputProblem = ({
     }
   }, [feedback])
 
-  const evaluateInput = (input: string, answer: string): boolean => {
+  const evaluateInput = (input: string | undefined, answer: string): boolean => {
```
rnathuji, line 71: I'm not sure why this signature needs to be changed. I don't think `input` will ever be `undefined` :man_shrugging: .

rnathuji (commented):
@P-Gill97 - I didn't look into the focus issue you mentioned during check-in (though I know you're still tinkering with that as well), but I wanted to share some other feedback that I had from a quick pass.

src/tests/kas.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    form: true
+  }
+  const ce = new ComputeEngine()
+
+  const consideredEqual = [
+    { expr1: ce.serialize(ce.parse('\\frac42')), expr2: ce.serialize(ce.parse('\\frac{4}{2}')) },
+    { expr1: 'x>4', expr2: '4<x' },
+    { expr1: 'x\\le4', expr2: '4\\gex' },
+    { expr1: '|-4|', expr2: '|4|' },
+    { expr1: '\\sqrt[3]{27}', expr2: '\\sqrt[3]{27}' },
```
rnathuji, line 16: Does this test case mean we can support and should add the nth root to the keyboard?

src/components/InputProblem.tsx
```diff
@@ -68,21 +69,45 @@ export const InputProblem = ({
...
     if (comparator.toLowerCase() === 'float') {
       return parseFloat(input) === parseFloat(answer)
     }
+    if (comparator.toLowerCase() === 'math') {
+      const ce = new ComputeEngine()
+
+      let parsedInput = parse(ce.serialize(ce.parse(input)))
+      let parsedAnswer = parse(ce.serialize(ce.parse(answer)))
+
+      // Sometimes compute engine produces an output that the KAS parse method does understand
```
rnathuji, line 86: Minor: I think you meant "does not understand" here.

src/components/InputProblem.tsx
```diff
@@ -68,21 +69,45 @@ export const InputProblem = ({
   }, [feedback])
 
   const evaluateInput = (input: string, answer: string): boolean => {
+    input = input.trim()
+    answer = answer.trim()
```
rnathuji, line 73: Just to avoid potential confusion when reading the code, it might be nicer to do something like:

```javascript
    const trimmedInput = input.trim()
    const trimmedAnswer = answer.trim()
```

and use those variables below.

src/components/Mathfield.tsx
```diff
@@ -64,6 +64,7 @@ export const Mathfield = ({ className, disabled, onInput }: MathfieldProps): JSX
         class={className}
         ref={mathfieldRef}
         onInput={onInput}
+        tabIndex={disabled ? '-1' : '0'}
```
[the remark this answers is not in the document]
rnathuji: Good catch!

src/tests/kas.test.ts
rnathuji: Yeah, you can go ahead and do that by updating the layout definition [here](https://github.com/openstax/k12-apps-raise/blob/592125cf97a1f81253061f1ff4603f12ce03d927/src/components/Mathfield.tsx#L6).

MReyna12: When we were originally trying to add the nth root key to the keyboard it was displaying oddly on the keyboard and in the input field. I couldn't figure out how the mathlive demo keyboard was displaying the nth root key properly, but that might be a place to look again if you run into the same issue.

rnathuji: Just a thought, but if it has to do with fonts, it could be because we do [this](https://github.com/openstax/k12-apps-raise/blob/592125cf97a1f81253061f1ff4603f12ce03d927/src/components/Mathfield.tsx#L4) while their demos probably use their fonts.

rnathuji: @P-Gill97 - Can you try setting that variable to `https://unpkg.com/mathlive/dist/fonts/` to see if it helps. I noticed [here](https://cortexjs.io/mathlive/guides/customizing/#fonts) it specifically says you need to user their fonts :joy: . If it works, we should confirm whether that URL is okay to use (it looks like it's a CDN :man_shrugging: ).

rnathuji: Just confirming that it's a reliable URL that's okay for "production". I'm not familiar with `unpkg.com`, but from a cursory look it seems like it would be okay. Just wanted another pair of 👀 to confirm. Otherwise we can figure out to host it somewhere as well 🤷‍♂️ .

src/components/Mathfield.tsx
```diff
@@ -1,13 +1,13 @@
...
 
-MathfieldElement.fontsDirectory = ''
+MathfieldElement.fontsDirectory = 'https://unpkg.com/mathlive/dist/fonts/'
 MathfieldElement.soundsDirectory = null
 window.mathVirtualKeyboard.layouts = {
   rows: [
     ['1', '2', '3', '4', '5', '6', '7', '8', '9', '0'],
     [
-      '\\sqrt{#0}', '#@^{#?}', '+', '-', '\\times', '\\frac{#@}{#?}', '|#0|', '(', ')'
+      '\\sqrt{#0}', '\\sqrt[#1]{#0}', '#@^{#?}', '+', '-', '\\times', '\\frac{#@}{#?}', '|#0|', '(', ')'
```
rnathuji, line 10: I know @MReyna12 and I discussed this a bit in a previous PR, but I'm having trouble understanding [this documentation](https://cortexjs.io/mathlive/guides/virtual-keyboards/#placeholder-tokens) as it relates to the differences between things like:
* `'\\sqrt[#1]{#0}'`
* `'\\sqrt[#@]{#?}'`
* ...

I see in the keyboard the latter creates a blue square vs an empty one, but experimentally I can't seem to figure out the implication from a user perspective. I wanted to check if you may have any thoughts / ideas (otherwise we can leave it as-is for now and revisit if we ever find an issue with it :man_shrugging: )

rnathuji (commented):
@P-Gill97 - I just have one final small comment (contradicting my previous self :joy: ) and a quick question.

rnathuji (approved):

PR #103, opened 2023-08-01
Update content version for AY23 and pin SU23 courses

rnathuji:


rnathuji:
@mjoss17 - Requesting your review on this one to confirm the final content version.

mjoss17 (approved):

PR #104, opened 2023-08-07
Update user input block

MReyna12:


MReyna12:
While playing around with the `<textarea>` I realized that if I apply the `.os-disabled` class to `<textarea>` then the resizing ability and scrollbar features are disabled due to `.os-disabled` stopping `pointer-events`. Previously we wanted users to not be able to interact with any of the interactives once in a disabled state so `pointer-events` were removed. However, users should still be able to resize the `<textarea>` and scroll the same even once they submit their answer or they may not be able to review their full answer post submission if they have not previously resized the `<textarea>`.

Instead of using `.os-disabled` to remove the box-shadow I added a "disabled" style (`.os-textarea-disabled`) to be used for `<textarea>`. I thought about separating removal of box-shadow and pointer-events from `.os-disabled`, but this would require adjusting the classes throughout the other interactives (essentially all interactives using `.os-disabled` would need `.os-no-box-shadow` and `.os-no-pointer-events` and `<textarea>` would only receive `.os-no-box-shadow`). I don't have a problem doing the aforementioned adjustments but wanted to get other opinions.

rnathuji (approved):
Good catch on the undesirable behavior of `.os-disabled`, @MReyna12 !

PR #106, opened 2023-08-21
Update tooltipify

P-Gill97:


rnathuji:
@P-Gill97 - Ah, I see. Your approach expects the caller to pass the data creating the need for a "global" in `main.tsx`. While I think that could work, I suspect it would force a bunch of stuff to wait on an async call to complete, and then create the need for dealing with what to do if it fails. Another approach could keep things in the same file but use a `Promise` that multiple async callers to `tooltipify` can wait on if needed (this is kind of what we do [here](https://github.com/openstax/k12-apps-raise/blob/92b36ebf89c2e78a37bb8e3b91269fb57d46713e/src/lib/events.ts#L103C69-L110)). 

That being said, I'm less concerned about not optimizing this if it's too painful / messy since:
* As opposed to the case of calling Moodle in the example above, this network request is for static content which the browser should cache. So, in the typical path, even if a page ends up calling `tooltipify` multiple times, they should get served from there.
* Glancing at our content, even when pages have multiple tooltips, from a rough sampling they tend to be in the same "block". So, really, if we do a network request inside of `tooltipify()`, I think the lowest hanging fruit is to ensure we only bother with it if there are `tooltipItems` (otherwise return early). That way the common case will never issue the request anyways.

Thoughts?

rnathuji:
@P-Gill97 - I think that generally seems aligned with my previous comments if you want to throw it into a PR for more specific feedback. A few notes that come to mind from a cursory glance at the above:

* I wouldn't use a `try / catch` to eat the `Error` in your `getGlossaryData` helper function. Instead, let that flow through so the caller to `tooltipify` can do what it wants with it (probably we'll just log there for now, but it allows callers to deal with error behavior vs hiding it)
* IDK that using `null` buys you much over `undefined` in your type definitions other than creating the need to then explicitly assign `null` 🤷‍♂️ 
* I don't think you need to create a manual `Promise` since `fetch` is going to give you one anyways. You could do something like `fetchPromise = fetchPromise ?? fetch(request)` if you were to define something like `let fetchPromise: Promise<Response>`.

rnathuji (approved):

PR #107, opened 2023-08-21
Update variants for Fall AY2023

mjoss17:


rnathuji:
@mjoss17 - This PR changes the wrong file. Can you revert the commit and modify `variant-mappings.json` instead of `content-versions.json`?

data/content-versions.json
```diff
@@ -2,116 +2,68 @@
...
+      "308": "animate-lomp-variant-b",
+      "309": "animate-lomp-variant-a",
+      "310": "animate-lomp-variant-a",
+      "311": "animate-lomp-variant-a",
+      "312": "animate-lomp-variant-a",
+      "313": "animate-lomp-variant-a",
+      "314": "animate-lomp-variant-b",
+      "315": "animate-lomp-variant-a",
+      "316": "animate-lomp-variant-a",
+      "317": "animate-lomp-variant-b"
```
rnathuji, line 66: Perhaps this is intentional, but this isn't currently mapping all of the existing AY23 courses. Also, perhaps we shouldn't map id `256` since that is a canary course.

data/variant-mappings.json
```diff
@@ -1,46 +1,68 @@
...
-      "193": "mm",
-      "194": "mm",
-      "202": "mm",
-      "208": "mm",
-      "210": "mm",
-      "211": "mm",
-      "214": "mm",
-      "230": "mm",
-      "232": "mm"
+      "256": "animate-lomp-variant-a",
```
rnathuji, line 4: Check my 👀 , but these variant names seem inconsistent with the content repo. It looks like the latter is using `animate-lomp-cameo-a` and `animate-lomp-cameo-b`.

rnathuji (approved):

PR #110, opened 2023-08-29
Update validation errors and related styling

MReyna12:


src/components/UserInputBlock.tsx
```diff
@@ -29,27 +29,34 @@ interface InputFormValues {
...
 
-  const handleSubmit = async (values: InputFormValues): Promise<void> => {
-    setResponseSubmitted(true)
+  const handleSubmit = async (values: InputFormValues, { setFieldError }: FormikHelpers<InputFormValues>): Promise<void> => {
+    if (values.response.trim() === '') {
+      setFieldError('response', NON_EMPTY_VALUE_ERROR)
+    } else {
+      setResponseSubmitted(true)
+    }
+
```
rnathuji, line 59: This is problematic since we continue on if there is an error and do things like fire events and invoke the `onInputSubmitted` handler. This would be analogous to unintentionally burning an attempt on the input problem component. One way to address this would be to use the `return` early pattern similar to what we're doing in the `InputProblem.tsx`:

```javascript
      return
    }

    setResponseSubmitted(true)
```

MReyna12: Thanks for catching this! I forgot to bring over the early return from the conditional logic used in `InputProblem.tsx`. I've updated accordingly.

src/components/UserInputBlock.tsx
```diff
@@ -29,27 +29,35 @@ interface InputFormValues {
...
 export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent, buttonText, contentId, onInputSubmitted }: UserInputBlockProps): JSX.Element => {
   const [responseSubmitted, setResponseSubmitted] = useState(false)
   const contentLoadedContext = useContext(ContentLoadedContext)
+  const NON_EMPTY_VALUE_ERROR = 'Please provide valid input'
 
   const schema = Yup.object({
     response: Yup.string()
-      .trim()
-      .required('Please provide valid input')
       .max(MAX_CHARACTER_INPUT_BLOCK_LENGTH, 'Input is too long')
```
rnathuji, line 50: Minor: Good call creating / using `NON_EMPTY_VALUE_ERROR`  in lieu of the in-line string definition (I think it helps with maintainability / readability even if not re-used :man_shrugging: ). It might be nice to go ahead and also define `EXCEEDED_MAX_INPUT_ERROR` and use that here.

rnathuji (approved):
I had a minor comment on a string definition in my second pass but going ahead and approving.

PR #111, opened 2023-09-12
Implement persistor for userInputBlock

MReyna12:


src/components/UserInputBlock.tsx
```diff
@@ -51,12 +66,66 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
+
   const handleSubmit = async (values: InputFormValues, { setFieldError }: FormikHelpers<InputFormValues>): Promise<void> => {
     if (values.response.trim() === '') {
       setFieldError('response', NON_EMPTY_VALUE_ERROR)
       return
     }
 
+    try {
+      if (contentId !== undefined && persistor !== undefined) {
+        await persistor.put(contentId, JSON.stringify({ userResponse: values.response, responseSubmitted: true }))
```
rnathuji, line 114: If we want to use a type definition to enforce our desired schema for persisted data, we may want to do something like:

```javascript
        const newPersistedData: PersistedData = { userResponse: values.response, responseSubmitted: true }
        await persistor.put(contentId, JSON.stringify(newPersistedData))
```

src/components/UserInputBlock.tsx
```diff
@@ -51,12 +66,66 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
+          const parsedLocalStorage = JSON.parse(currentLocalStorageState)
+          setInteractiveState({
+            userResponse: parsedLocalStorage.userResponse,
+            responseSubmitted: parsedLocalStorage.responseSubmitted
+          })
+        }
+      }
+    } catch (error) {
+      setPersistorStatus(PersistorStatus.Failure)
+      console.error(error)
```
rnathuji, line 126: This log can go away since errors will be surfaced in the UI.

src/components/UserInputBlock.tsx
```diff
@@ -86,34 +155,56 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
       < div className='my-3 os-feedback-message ' ref={contentRefCallback} dangerouslySetInnerHTML={{ __html: ack }} />
       )
 
-  return (
-    <EventControlledContent waitForEvent={waitForEvent}>
+  if (interactiveState !== null && persistorStatus === PersistorStatus.Success) {
```
rnathuji, line 158: This is subjective / cosmetic, but what do you think about having the comment path be the default return at the end and other two conditional on state of the persistor `.get()`? Since this is the most complicated return, it would also reduce some indentation on it :man_shrugging: .

src/components/UserInputBlock.tsx
```diff
@@ -51,12 +66,66 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
+        if (currentLocalStorageState !== null) {
+          const parsedLocalStorage = JSON.parse(currentLocalStorageState)
+          setInteractiveState({
+            userResponse: parsedLocalStorage.userResponse,
+            responseSubmitted: parsedLocalStorage.responseSubmitted
+          })
+        }
+      }
+    } catch (error) {
+      setPersistorStatus(PersistorStatus.Failure)
```
rnathuji, line 125: I'm not sure this is the best way to handle the scenario where a `.put()` fails since it would effectively hide everything and the user can't proceed in any reasonable way. What about something like the following instead where we effectively treat being able to save the response as post-submit validation?:

```javascript
      setFieldError('response', 'Error saving response. Please try again.')
      return
```

rnathuji (commented):
@MReyna12 - Apologies for the not entirely self-consistent review comments as when I dug in a bit I realized I had some thoughts that might negate / alter other comments :joy: . A couple of high level contextual points for some of the feedback:

* I had a bunch of naming comments mostly because, even though this is a spike, I think getting some of those cleared up will help us establish shared mental models as to what's going on
* I'm thinking we may not actually need any component state to handle rehydration, and instead (in this case) it may just be we need to explicitly store `response` state. Either way, establishing a pattern we think makes the most sense will help us with consistency in ongoing spike work.

src/components/UserInputBlock.tsx
```diff
@@ -23,12 +24,24 @@ interface UserInputBlockProps {
...
   response: string
 }
 
+interface PersistorState {
+  userResponse: string
+  responseSubmitted: boolean
+}
+
+enum PersistorStatus {
+  Impersistent,
```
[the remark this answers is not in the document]
MReyna12: Unitialized has my vote.

src/lib/blocks.tsx
```diff
@@ -213,6 +214,15 @@ export const parseUserInputBlock = (element: HTMLElement): JSX.Element | null =>
...
   }
 
+  const browserPersistor: Persistor = {
+    get: async (key: string): Promise<string | null> => {
+      return localStorage.getItem(key)
+    },
+    put: async (key: string, value: string) => {
+      localStorage.setItem(key, value)
+    }
+  }
```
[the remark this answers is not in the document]
MReyna12: Agreed - when we originally spoke about passing `persistor={browserPersistor}` as a prop to `UserInputBlock` I misunderstood the placement of `browserPersistor`. It is now moved to `persistor.ts` and imported accordingly.

src/components/UserInputBlock.tsx
```diff
@@ -1,10 +1,11 @@
 import { EventControlledContent } from './EventControlledContent'
```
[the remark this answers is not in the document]
MReyna12: That makes complete sense, thanks.

src/components/UserInputBlock.tsx
MReyna12: Fully acknowledge it was not the best way to handle the failure of the `.put()`. Because this is a spike I did not put too much thought into error handling, but your implementation is simple enough that I'll keep it in mind moving forward. Thanks!

MReyna12:
All comments have been addressed in latest PR. Per your comment [here](https://github.com/openstax/k12-apps-raise/pull/111#discussion_r1324655008) and our discussion via Slack, I have updated the state management pattern to mirror what is shown in the aforementioned comment.

src/components/UserInputBlock.tsx
```diff
@@ -51,12 +66,45 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
+  const getPersistedState = async (): Promise<void> => {
+    try {
+      if (contentId === undefined || persistor === undefined) {
+        setPersistorGetStatus(PersistorGetStatus.Success)
+        return
+      }
+
+      const persistedState = await persistor.get(contentId)
+      if (persistedState !== null) {
+        const parsedLocalStorage = JSON.parse(persistedState)
```
rnathuji, line 78: Ah, missed this minor name nit to remove references to local. Maybe `parsedPersistedState`?

MReyna12: I told myself to update that and forgot to address after our huddle so no worries. Thanks for catching. Updated in latest commit.

rnathuji (approved):

PR #113, opened 2023-09-18
Create textheavyadjustedtable

MReyna12:


MReyna12:
A few notes for the reviewer regarding the acceptance criteria. Based on our discussion during engineering check-in + the note regarding naming of the new table class in the description, I believe the new criteria is:
1. There is a new table type defined as "os-raise-textheavyadjustedtable"
2. All columns of the table are not fixed, but rather the width adjusts to the content
3. All the table text is left aligned

src/styles/content.scss
```diff
@@ -228,6 +228,33 @@ table.os-raise-horizontaltable {
   }
 }
 
+table.os-raise-textheavyadjustedtable {
```
rnathuji, line 231: Would this be equivalent to something like the following?:

```scss
%os-raise-textheavytable-base {
  @extend %os-raise-basictable;

  width: 100%;

  th,
  td {
    text-align: left;
  }

  th {
    vertical-align: bottom;
  }

  td {
    vertical-align: top;
  }

  td:first-child {
    border-left: none;
  }
}

table.os-raise-textheavytable {
  @extend %os-raise-textheavytable-base;

  table-layout: fixed;
}

table.os-raise-textheavyadjustedtable {
  @extend %os-raise-textheavytable-base;

  table-layout: auto;
}
```

If so, would that be easier to maintain / read?

MReyna12: Yes it would - thanks for the suggestion.

src/styles/content.scss
```diff
@@ -228,6 +228,33 @@ table.os-raise-horizontaltable {
   }
```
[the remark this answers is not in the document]
MReyna12: Test case added and styles.md updated in latest commit.

docs/styles.md
```diff
@@ -4,6 +4,7 @@
   - [Styling Content for RAISE](#styling-content-for-raise)
 - Tables
   - [Text Heavy Tables](#text-heavy-table)
+  - [Adjustable Column Text Heavy Tables](#text-heavy-adjusted-table)
```
rnathuji, line 7: This anchor link doesn't seem to work. Maybe it needs to be `#adjustable-column-text-heavy-table` or we should update the title align with this anchor :man_shrugging:

rnathuji (approved):

PR #115, opened 2023-09-22
Add persistor to pset components

MReyna12:


MReyna12:
Notes:

1. For the multiselect interactive: In order to use an array inside of the useEffect dependency i wrapped it with `JSON.stringify()`. Per this [discussion](https://github.com/facebook/react/issues/14476#issuecomment-471199055) it looks like this is a viable option in our scenario, but I am open to another solution either outlined in the discussion or somewhere else.
2. All PSET interactives:
  a. Error handling on failed `put` - Currently, I am only using `console.error(err)`. Because the `put` occurs outside of the `handleSubmit`, I do not think I can use `setFieldError` in the same manner as `UserInputBlock.tsx`. Any ideas on how to handle a `put` failure?
  b. I believe we need `retriesAllowed` and `formDisabled`/`inputDisabled` as a dependency of the useEffect. I've also included `response` as a dependency, but `response` updating occurs when `retriesAllowed` and `formDisabled`/`inputDisabled` updates as well. Should we include `response` as a dependency or would `retriesAllowed` and `formDisabled`/`inputDisabled` be sufficient?

src/components/InputProblem.tsx
```diff
@@ -105,11 +121,54 @@ export const InputProblem = ({
...
+      setPersistorGetStatus(PersistorGetStatus.Success)
+    } catch (err) {
+      setPersistorGetStatus(PersistorGetStatus.Failure)
+    }
+  }
+
+  useEffect(() => {
+    setPersistedState().catch(() => { })
+    getPersistedState().catch(() => { })
+  }, [response, inputDisabled, retriesAllowed])
```
rnathuji, line 163: This is problematic because the `useEffect()` will do a get more than once, which we should never have to do. At the very least, I think we should separate that one out to ensure there is only ever one `.get()` call invoked by a component. It's also a bit odd because we do a `.put()` before we do a `.get()` (it only happens to work out I think because of the conditional `response !== ''` :man_shrugging: )

The use of a `.put()` seems to be a different pattern than `UserInputBlock` and has unfortunate consequences per [your comment](https://github.com/openstax/k12-apps-raise/pull/115#issuecomment-1731754344). Perhaps this was motivated to handle the clearing of feedback in `clearFeeback()` (maybe?), but I don't think that's necessary. See related comment in `handleSubmit()`.

src/components/InputProblem.tsx
```diff
@@ -105,11 +121,54 @@ export const InputProblem = ({
...
+
   const handleSubmit = async (values: InputFormValues, { setFieldError }: FormikHelpers<InputFormValues>): Promise<void> => {
```
rnathuji, line 165: While I think this submit handler is getting too complicated and we should see if we can't refactor / simplify in some way, from a functional standpoint I think it is very similar to what you had before but maintains the pattern that we "save" on submit:

```javascript
  const handleSubmit = async (values: InputFormValues, { setFieldError }: FormikHelpers<InputFormValues>): Promise<void> => {
    let correct = false
    let finalAttempt = false
    const attempt = retriesAllowed + 1

    const setPersistedState = async (persistorData: PersistorData): Promise<void> => {
      if (contentId === undefined || persistor === undefined) {
        return
      }
      await persistor.put(contentId, JSON.stringify(persistorData))
    }

      if ((comparator.toLowerCase() === 'integer') || (comparator.toLowerCase() === 'float')) {
        setFieldError('response', NUMERIC_INPUT_ERROR)
      } else {
      }
      return
    }

    if (evaluateInput(values.response, solution)) {
      correct = true
      try {
        await setPersistedState({ userResponse: values.response, inputDisabled: true, retriesAllowed, feedback: correctResponse })
      } catch (error) {
        setFieldError('response', 'Error saving response. Please try again.')
        return
      }

      setFeedback(correctResponse)
      solvedCallback()
      setInputDisabled(true)
    } else if (retryLimit === 0 || retriesAllowed !== retryLimit) {
      const feedback = determineFeedback(values.response, encourageResponse, answerResponses, evaluateInput)
      try {
        await setPersistedState({ userResponse: values.response, inputDisabled, retriesAllowed: retriesAllowed + 1, feedback })
      } catch (error) {
        setFieldError('response', 'Error saving response. Please try again.')
        return
      }

      setRetriesAllowed(currRetries => currRetries + 1)
      setFeedback(feedback)
      allowedRetryCallback()
    } else {
      try {
        await setPersistedState({ userResponse: values.response, inputDisabled: true, retriesAllowed: retriesAllowed + 1, feedback: attemptsExhaustedResponse })
      } catch (error) {
        setFieldError('response', 'Error saving response. Please try again.')
        return
      }

      setRetriesAllowed(currRetries => currRetries + 1)
      setFeedback(attemptsExhaustedResponse)
      exhaustedCallback()
      setInputDisabled(true)
      finalAttempt = true
    }

    setResponse(values.response)

    if (onProblemAttempt !== undefined) {
      onProblemAttempt(
        values.response,
        correct,
        attempt,
        finalAttempt,
        contentId
      )
    }
  }
```

Admittedly, we don't "save" clearing the feedback, but I think that's fair since it's more of an ephemeral state than the "persisted" state in my mind. Said differently, I think the "persisted" contract should be strictly on an attempt level, and not the "`onChange`" level (both for simplicity and to minimize server load impacts of writes).

src/components/InputProblem.tsx
```diff
@@ -105,11 +121,54 @@ export const InputProblem = ({
     return trimmedInput.toLowerCase() === trimmedAnswer.toLowerCase()
   }
 
+  const setPersistedState = async (): Promise<void> => {
```
rnathuji, line 124: I think this can be pulled out and moved into a helper function in `handleSubmit()`.

rnathuji (commented):
@MReyna12 - This is a partial review of just `ProblemSetBlock` and `InputProblem`. The reason being, I think depending on thoughts there I think it would impact the other components in a similar way. See what you think about my thoughts, but I think they would address both of the items you raised [here](https://github.com/openstax/k12-apps-raise/pull/115#issuecomment-1731754344):

* Item 1: It would remove the pattern of using state as in the `useEffect` dependencies altogether
* Item 2: The `.put()` would get moved into `handleSubmit()` so we could employ the same error handling pattern we have in `UserInputBlock`.

src/components/InputProblem.tsx
```diff
@@ -105,11 +121,54 @@ export const InputProblem = ({
...
+        return
+      }
+
+      const persistedState = await persistor.get(contentId)
+      if (persistedState !== null) {
+        const parsedPersistedState = JSON.parse(persistedState)
+        setResponse(parsedPersistedState.userResponse)
+        setInputDisabled(parsedPersistedState.inputDisabled)
+        setRetriesAllowed(parsedPersistedState.retriesAllowed)
+        setFeedback(parsedPersistedState.feedback)
```
rnathuji, line 152: This seems unfortunate to me in that we're having to store state that we could, in theory, recompute and could otherwise be of an unknown size (e.g. arbitrary HTML).  How difficult / annoying would it be to avoid that?

MReyna12:
@rnathuji - My latest commit addresses your comments. I am still trying to see if there is anything that can be done in the submit handler to not effectively write the same try/catch statement three times (unless you think it is not necessary). Once these updates get approval I will make similar updates to the other components.

src/components/InputProblem.tsx
```diff
@@ -10,10 +10,12 @@ import { Mathfield } from './Mathfield'
 import { type MathfieldElement } from 'mathlive'
 import { parse, compare } from '@khanacademy/kas'
 import { ComputeEngine } from '@cortex-js/compute-engine'
+import { type Persistor } from '../lib/persistor'
 export const MAX_CHARACTER_INPUT_PROBLEM_LENGTH = 500
 
 interface InputProblemProps extends BaseProblemProps {
   comparator: string
+  persistor?: Persistor
```
rnathuji, line 18: Instead of defining this here, you can add it to `BaseProblemProps` in `ProblemSetBlock.tsx` so it gets included for all of the problem types.

src/components/InputProblem.tsx
```diff
@@ -121,19 +122,17 @@ export const InputProblem = ({
...
-        return
-      }
-
-      if (response !== '' || inputDisabled || retriesAllowed !== 0) {
-        const newPersistedData: PersistorData = { userResponse: response, inputDisabled, retriesAllowed, feedback }
-        await persistor.put(contentId, JSON.stringify(newPersistedData))
-      }
-    } catch (err) {
-      console.error(err)
+  const handleFeedback = (userResponse: string, userAttempts: number): string => {
```
rnathuji, line 125: Just a thought: Right now we end up calling `evaluateInput` multiple times on a submit. The function is pretty light except for `math` problems where we invoke `ce` / `kas`, but I wanted to note it in case there's an alternative implementation that might avoid that :man_shrugging: .

src/components/InputProblem.tsx
```diff
@@ -105,11 +122,56 @@ export const InputProblem = ({
...
+    if (evaluateInput(userResponse, solution)) {
+      setFeedback(correctResponse)
+    } else if (retryLimit === 0 || userAttempts !== retryLimit + 1) {
+      const attemptsRemainingResponse = determineFeedback(userResponse, encourageResponse, answerResponses, evaluateInput)
+      setFeedback(attemptsRemainingResponse)
+    } else {
+      setFeedback(attemptsExhaustedResponse)
+    }
+
+    return feedback
```
rnathuji, line 135: I'm not sure that returning this is doing what you intend in this function since `feedback` may not necessarily be what you end up invoking `setFeedback` with until after the hook runs. If you want to return the feedback string (though, no callers currently use it so you could alternatively make this function return `void`), perhaps use an early `return` in each conditional?

src/components/InputProblem.tsx
```diff
@@ -105,11 +122,56 @@ export const InputProblem = ({
     return trimmedInput.toLowerCase() === trimmedAnswer.toLowerCase()
   }
 
+  const handleFeedback = (userResponse: string, userAttempts: number): string => {
+    if (evaluateInput(userResponse, solution)) {
+      setFeedback(correctResponse)
+    } else if (retryLimit === 0 || userAttempts !== retryLimit + 1) {
```
rnathuji, line 128: Should this not be `(retryLimit === 0 || retriesAllowed !== retryLimit)`? If so, does `userAttempts` even need to get passed here?

rnathuji (commented):
@MReyna12 - I think the changes to `InputProblem` in general look good and can be replicated across. I  don't see an  obvious way to avoid the multiple `try / catch` statements, and perhaps that's something we can just defer to a future "real" implementation since what you have does address the key pieces. I did have a few comments on the new `handleFeedback()`.

src/components/InputProblem.tsx
rnathuji: Oh -- wait, I think you have to pass it for the `getPersistedState` path invocation. I'm still not uncertain there's not a potential off-by-one error being introduced, and it makes me nervous that it's not obvious either way :sweat_smile: .

MReyna12: That's right. If I were to do `retriesAllowed !== retryLimit` then when refreshing on the attempt before the last attempt you will get the `attemptsExhaustedResponse` even though the user should have one more attempt. Would we want to forego displaying the feedback on refresh? Then the user would only see the feedback after they submit.

src/components/InputProblem.tsx
MReyna12: The `return` was part of me initially testing something out in the function. I'll remove and early return conditionals.

src/components/InputProblem.tsx
MReyna12: Initial thought was we want to compare the the `userResponse` with the `solution` and `evaluateInput` handles the various `comparator` types so I opted for `evaluateInput`. 

I'll give this one some thought to see if there is an alternative.

src/components/InputProblem.tsx
rnathuji: Ah, what about if in `handleSubmit()` we always invoke `handleFeedback(values.response, retriesAllowed)`. In fact, in that case we can just pull it out and do it with `setResponse`:

```javascript
    handleFeedback(values.response, retriesAllowed)
    setResponse(values.response)
```

...then `handleFeedback` can use the same conditional logic:

```javascript
  const handleFeedback = (userResponse: string, userAttempts: number): void => {
    if (evaluateInput(userResponse, solution)) {
      setFeedback(correctResponse)
    } else if (retryLimit === 0 || userAttempts !== retryLimit) {
      setFeedback(determineFeedback(userResponse, encourageResponse, answerResponses, evaluateInput))
    } else {
      setFeedback(attemptsExhaustedResponse)
    }
  }
```

The wrinkle is really on the `getPersistedState` path, but I think we can handle that by invoking the method slightly differently there:

```javascript
handleFeedback(parsedPersistedState.userResponse, parsedPersistedState.retriesAllowed > 0 ? parsedPersistedState.retriesAllowed - 1 : 0)
```

The idea is that, in essence, on the rehydration path to invoke the `handleFeedback` the way it was originally, we need to "rewind by one". The ternary checks for 0 as I suspect that's what we'll reset to eventually when we implement [k12-493](https://github.com/openstax/k12/issues/493). 

Does that work, or is there some bug it now introduces? :thinking:

MReyna12: It appears to work for `InputProblem.tsx`. I'll update the logic accordingly across the other components (except for `UserInputBlock.tsx`).

I understand how this refactor essentially uses the same version of `retriesAllowed` without the persistor, but how is adding 1 to `retryLimit` paired with the updated `retriesAllowed` different from subtracting 1 from `retriesAllowed`? Is this the preferred method because we know functionally it should operate similarly to the non-persistor version of the component?

rnathuji: I think it's just confusing to have:

1) (Needlessly) different invocations to `handleFeedback()` in `handleSubmit()`. It makes a reader think something special is happening, but there's really not. In general, it seems much cleaner (subjective) to let the `getPersistedState` do "something special" vs muddy up the `handleSubmit()` path. Said differently, it's overly confusing to do a '+1' and then (the algebraic equivalent of) '-1' in the same code path. 
2) Since, in this incarnation, we're reproducing the same `if / else / ...` statement in `handleSubmit()` and `handleFeedback()`, it's very confusing to a reader that there is a different conditional statement IMHO. My concern is it could lead to "off by one" errors, and it definitely wasn't clear to me when reviewing that your change hadn't introduced one.

MReyna12: Thanks for the explanation! I see how my implementation provides less clarity and could lead the reader down an unintended path.

src/components/InputProblem.tsx
MReyna12: Per our discussion below, `handleFeedback` conditional logic now mirrors the conditional logic in `handleSubmit`. You can ignore my previous comment.

src/components/DropdownProblem.tsx
```diff
@@ -66,28 +80,92 @@ export const DropdownProblem = ({
...
+
       setRetriesAllowed((currRetries) => currRetries + 1)
-      setFeedback(attemptsExhaustedResponse)
       exhaustedCallback()
       setFormDisabled(true)
       finalAttempt = true
     }
 
+    handleFeedback(values.response, retriesAllowed)
+    setResponse(values.response)
```
rnathuji, line 167: I think this goes back to the naming issues you suggested in [this comment](https://github.com/openstax/k12/issues/513#issuecomment-1737821810), but I don't think this `setResponse` is really necessary. That would have probably been more clear if we had called things:

```javascript
const [initialResponse, setInitialResponse] = useState('')
```

Would :point_up: be a useful change to make across all of the components (including `UserInputBlock`) for clarity / to not exacerbate the confusion around "response" in the code in this branch?

src/components/InputProblem.tsx
rnathuji: Let's punt on this and revisit in a future "real" implementation.

src/components/DropdownProblem.tsx
MReyna12: Agreed that it is not necessary - I believe I pulled that from some the suggested refactor to my original `handleSubmit`. We should only need to `setResponse` in the `get`. 

I'll update `response` to `initialResponse` across the components, including `UserInputBlock`.

rnathuji (approved):

PR #118, opened 2023-10-09
Improve code readability

MReyna12:


src/components/MultipleChoiceProblem.tsx
```diff
@@ -1,5 +1,5 @@
 import type { BaseProblemProps } from './ProblemSetBlock'
-import { determineFeedback } from '../lib/problems'
+import { determineFeedback, retriesRemaining } from '../lib/problems'
 import { useCallback, useState } from 'react'
```
rnathuji, line 3: Putting this here because I can't comment on it directly: It's a judgement call either way, but I wanted to confirm that you didn't change `buildClassName` to replace `solution: string` with `correct: boolean` because the addition of the `userResponseCorrect` state where it otherwise wouldn't get used during rendering wasn't a worthwhile tradeoff?

There is a subtle problem we have in `buildClassName` where we're then effectively reproducing the operator used to compare values in `evaluateInput()`, but in this case it's just `===` so the risk exists in practice only if that ever changes which seems pretty unlikely for this problem type :man_shrugging: (even in this worst case scenario, it would surface a styling wart vs a functional one).

Very verbose way of saying: I agree with your thought process if :point_up: is the case, but I wanted to confirm.

src/components/MultiselectProblem.tsx
```diff
@@ -139,30 +149,22 @@ export const MultiselectProblem = ({
     const attempt = retriesAllowed + 1
     if (compareForm(values.response, solutionArray)) {
```
rnathuji, line 150: Only because it confused me when looking at this code, what would you think about doing the following:

1) Rename the current `compareForm` to `evaluateInput` and use that here (this would make the naming / usage consistent with other problems)
2) Move the current `evaluateInput` inside of the `handleFeedback` and call it `comparator`. So something like:

```javascript
  const handleFeedback = (userResponse: string[], correct: boolean, userAttempts: number): void => {
    const comparator = (input: string[], answer: string): boolean => {
      return evaluateInput(input, JSON.parse(answer))
    }
```

Then pass that to `determineFeedback`:

```javascript
setFeedback(determineFeedback(userResponse, encourageResponse, answerResponses, comparator))
```

src/components/MultipleChoiceProblem.tsx
MReyna12: That was my thought process - since `userResponseCorrect` wouldn't be used during rendering like in `InputProblem` and `DropdownProblem`, I decided to forego replacing `solution: string` with `correct: boolean`.

One question - `evaluateInput()` takes in `values.response` and `solution`. We are not comparing `solution` to `response` in `buildClassName`. `val` in `buildClassName` represents each of the MC answer choices. Would the alternative be to drop `solution === val` for `correct` where `correct` is `true` if `evaluateInput(values.response, solution)`?

rnathuji: Yeah, I think so as far as passing in the function to avoid assuming the comparison logic.

PR #119, opened 2023-10-10
Implement moodle persistor in moodleApi and persistor

P-Gill97:


src/lib/env.ts
```diff
@@ -1,6 +1,7 @@
 export const ENV = {
   OS_RAISE_CONTENT_URL_PREFIX: import.meta.env.MODE === 'production' ? 'https://k12.openstax.org/contents/raise' : 'http://localhost:8800/contents',
   OS_RAISE_EVENTSAPI_URL_MAP: {},
+  OS_RAISE_ENABLE_MOODLE_PERSISTOR_MAP: {},
```
rnathuji, line 4: Can we slightly modify this to be a list? I think it will make things simpler:

```javascript:
OS_RAISE_ENABLE_MOODLE_PERSISTOR_HOSTS: [] as string[],
```

src/lib/env.ts
```diff
@@ -13,6 +14,9 @@ if (import.meta.env.MODE === 'production') {
   ENV.OS_RAISE_EVENTSAPI_URL_MAP = {
     'localhost:8000': 'http://localhost:8888'
   }
+  ENV.OS_RAISE_ENABLE_MOODLE_PERSISTOR_MAP = {
```
rnathuji, line 17: Based on comment above, we can make this:

```javascript
  ENV.OS_RAISE_ENABLE_MOODLE_PERSISTOR_HOSTS = [
    'localhost:8000'
  ]
```

src/lib/persistor.ts
```diff
@@ -1,13 +1,26 @@
...
-export const browserPersistor: Persistor = {
-  get: async (key: string): Promise<string | null> => {
-    return localStorage.getItem(key)
-  },
-  put: async (key: string, value: string) => {
-    localStorage.setItem(key, value)
+export const getMoodlePersistor = (): Persistor | undefined => {
+  const { courseId } = getCurrentContext()
+
+  if (courseId === undefined || window.M === undefined) {
```
rnathuji, line 12: We can add the following to use the `ENV` list to enforce disabling:

```javascript
  if (courseId === undefined || window.M === undefined || host === undefined || !ENV.OS_RAISE_ENABLE_MOODLE_PERSISTOR_HOSTS.includes(host)) {
    return undefined
  }
```

The `host` used :point_up: comes from the context.

src/lib/persistor.ts
```diff
@@ -1,13 +1,26 @@
...
 }
 
-export const browserPersistor: Persistor = {
-  get: async (key: string): Promise<string | null> => {
-    return localStorage.getItem(key)
-  },
-  put: async (key: string, value: string) => {
-    localStorage.setItem(key, value)
+export const getMoodlePersistor = (): Persistor | undefined => {
+  const { courseId } = getCurrentContext()
```
rnathuji, line 10: I know this is something we'd need to reflect in `main` later (e.g. to modify `events.ts` to use the data from context), but just to test / set the example here even though it's a spike branch, I'm wondering if it isn't better to just modify `getCurrentContext` so accesses to any of the magical moodle `window.M` variables is done there. Something like:

```javascript
export const getCurrentContext = (): { courseId: number | undefined, host: string | undefined, moodleWWWRoot: string | undefined, moodleSessKey: string | undefined } => {
  const courseId = window.M?.cfg.courseId
  const host = window.location.host
  const moodleWWWRoot = window.M?.cfg.wwwroot
  const moodleSessKey = window.M?.cfg.sesskey

  return {
    courseId,
    host,
    moodleWWWRoot,
    moodleSessKey
  }
```

I think doing this will help us encapsulate things a bit in terms of isolating where assumptions on "being in Moodle" get made.

rnathuji (commented):
Nice, things seem to have come together without any unexpected issues! 

I only had a few minor code comments on the ENV flag and consolidating access to `window.M` to live in `getCurrentContext()`.

rnathuji (approved):

PR #120, opened 2023-10-11
Add reset for state and UI

MReyna12:


src/components/UserInputBlock.tsx
```diff
@@ -176,8 +196,21 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
-              <button type="submit" disabled={isSubmitting || responseSubmitted} className="os-btn btn-outline-primary">{buttonText}</button>
+              <div className='os-text-center mt-4 os-flex os-justify-space-evenly'>
+                <button type="submit" disabled={isSubmitting || responseSubmitted} className="os-btn btn-outline-primary">{buttonText}</button>
+                {
+                  (persistor !== undefined) &&
+                  <button
+                    type="button"
+                    onClick={(): void => {
+                      void resetPersistedState(setFieldError, setFieldValue)
+                      void setTouched({ response: true }, false)
```
rnathuji, line 207: Clarification question: What's the behavioral difference if the `false` isn't passed and validation is enabled? Playing around I couldn't quite figure it out. I expected it would give a validation error without it when we reset to `''`, but that doesn't seem to be the case :man_shrugging: .

src/components/DropdownProblem.tsx
```diff
@@ -81,6 +81,10 @@ export const DropdownProblem = ({
   }
 
   const handleFeedback = (userResponse: string, userAttempts: number): void => {
+    if (userResponse === '') {
+      return
+    }
+
```
rnathuji, line 87: Clarification question for all of the problems: What happens if this check is removed? (I tested it on MC, but nothing obvious broke :man_shrugging: )

src/components/UserInputBlock.tsx
```diff
@@ -89,6 +89,26 @@ export const UserInputBlock = ({ content, prompt, ack, waitForEvent, fireEvent,
...
+    setFieldValue: (field: string, value: any, shouldValidate?: boolean | undefined) => Promise<void | FormikErrors<InputFormValues>>
+  ): Promise<void> => {
+    try {
+      if (contentId === undefined || persistor === undefined) {
+        return
+      }
+
+      const newPersistedData: PersistorData = { userResponse: '', responseSubmitted: false }
+      await persistor.put(contentId, JSON.stringify(newPersistedData))
+      setInitialResponse('')
```
rnathuji, line 104: Clarification question: Is this necessary since we're not using the `onReset` prop?

rnathuji (commented):
@MReyna12 - As it relates to [this comment](https://github.com/openstax/k12/issues/493#issuecomment-1761943469), I don't think what you've done here is confusing. I did have a few clarification questions as far as some of the calls / flags.

src/components/UserInputBlock.tsx
MReyna12: I didn't notice a behavioral difference either, but since we were resetting to our default value I didn't see the point in validating the default value so I passed `false`. Maybe I am misunderstanding the point of validation?

src/components/DropdownProblem.tsx
MReyna12: Feedback will show if the user submits an answer, resets, and then refreshes the page. Without this you hit the second conditional below. I've included a video example of the behavior without the early return.

```js
if (evaluateInput(userResponse, solution)) {
      setFeedback(correctResponse)
    } else if (retryLimit === 0 || userAttempts !== retryLimit) {
      setFeedback(determineFeedback(userResponse, encourageResponse, answerResponses, evaluateInput))
    } else {
      setFeedback(attemptsExhaustedResponse)
    }
```

https://github.com/openstax/k12-apps-raise/assets/37000585/26b0021d-e4b9-4b4b-bc4e-7806d044f3a7

src/components/UserInputBlock.tsx
MReyna12: Are you referring to all three lines? Without making the `put` the state previously stored via the `get` is displayed on refresh. I actually do not notice a UI difference if I remove line 104.

MReyna12: After further testing across the psets it looks like I do not need to set the response in `resetPersistedState`. `setFieldValue()` clears the input and if a user refreshes after resetting then when the `get` is made the UI would render with the empty response value.

src/components/DropdownProblem.tsx
rnathuji: Ah -- right, of course. Thanks!

src/components/UserInputBlock.tsx
rnathuji: Ah, so it's just a bit of efficiency to avoid doing unnecessary work. :+1:

src/components/UserInputBlock.tsx
rnathuji: Is it worth removing them, then (if for no other reason then so we don't get confused when referencing this code in the future)?

MReyna12: Yes. I removed them and I'll push the update soon.

rnathuji (approved):

PR #122, opened 2023-10-18
Add handler to maintain event firing

MReyna12:


src/components/DropdownProblem.tsx
```diff
@@ -80,17 +80,20 @@ export const DropdownProblem = ({
...
   }
 
-  const handleFeedback = (userResponse: string, userAttempts: number): void => {
+  const handleProblemResult = (correct: boolean, userAttempts: number, userResponse: string): void => {
     if (userResponse === '') {
       return
     }
 
-    if (evaluateInput(userResponse, solution)) {
+    if (correct) {
```
rnathuji, line 88: I think what you have here aligns with how the event firing logic works today in `ProblemSetBlock.tsx`, but I think it's slightly incomplete. For each problem we store the following at the pset level:

```javascript
interface ProblemResult {
  solved: boolean
  exhausted: boolean
  attempts: number
}
```

I believe the handling here will sufficiently "replay" state for `solved` and `exhausted`, but the `attempts` count would be incorrect. It just so happens we don't use that today so there's no perceivable error. How hairy would it be to address the gap (assuming I'm not mistaken)? Would it perhaps amount to calling `allowedRetryCallback()` `parsedPersistedState.retriesAllowed - 1` times (if it's greater than 0) in `getPersistedState` prior to invoking `handleProblemResult`?

rnathuji: As a note: If you use the React extension to inspect component state, I noticed that with `React.StrictMode` enabled you get non-deterministic values for `attempts` after the persistor loads. Presumably this has to do with the way StrictMode will run things multiple times, but just a heads up :joy: .

MReyna12: Per a huddle with @rnathuji - We've decided, for now, to not revise the code to properly account for `attempts`. The reason being is at this time we do not make use of `attempts` and the incorrect number of `attempts` does not currently impact the user. 

We want to discuss cleaning up `ProblemSetBlock.tsx` to remove what is currently dead code related to `attempts`. The benefit to doing so would allow us to maintain less code and reduce the likelihood of errors / confusion for the dev team. However, before this decision is made we should discuss with PDM to ensure this is not something that they may want to keep around.

rnathuji: I've added :point_up: to the [list of things we'll need to consider](https://docs.google.com/presentation/d/1ehdDtPN8iQgKfWPXd3UWK_4P1N5fUQWrZrJiuSjsGQI/edit#slide=id.g24906625bce_0_20) so we don't lose track of it. We can wrap any corresponding changes to `ProblemSetBlock.tsx` into implementation design / requirements accordingly.

rnathuji (approved):

PR #123, opened 2023-10-19
Add unfire event and handler for the same

MReyna12:


MReyna12:
I am open to suggestions on `problemResultsResetValue` in `ProblemSetBlock.tsx`.

src/components/ProblemSetBlock.tsx
```diff
@@ -99,6 +101,7 @@ export const ProblemSetBlock = ({ waitForEvent, fireSuccessEvent, fireLearningOp
     // are either solved or exhausted
     const solvedCount = Array.from(problemResults.values()).reduce((acc, prob) => (prob.solved ? acc + 1 : acc), 0)
     const exhaustedCount = Array.from(problemResults.values()).reduce((acc, prob) => (prob.exhausted ? acc + 1 : acc), 0)
+    const problemResultsResetValue = Array.from(problemResults.values())[0].reset
```
rnathuji, line 104: This is currently hard coded to only work if the first problem is reset and doesn't work with psets that have multiple problems if you reset any of the rest. I think the following reducer will result in `true` if any of the problems are `true` (but, you should confirm / check my logic :sweat_smile: ):

```javascript
const problemResultsResetValue = Array.from(problemResults.values()).reduce((acc, prob) => (prob.reset ? true : acc), false)
```

rnathuji (commented):
The problem set unfire approach seemed to come together nicely with the `resetCallback` addition :tada: . I just noticed one minor issue in the implementation when reviewing.

src/components/ProblemSetBlock.tsx
MReyna12: Your reducer does operate as I would expect it to / results in `true` upon pressing the `reset` button. All of my "test" examples use one problem per pset. I'll adjust then moving forward to hopefully catch that undesired behavior. Thanks for the rewrite!

rnathuji (approved):

PR #129, opened 2023-10-31
Add responsive img class and update markdown

MReyna12:


src/styles/content.scss
```diff
@@ -275,4 +275,9 @@ table.os-raise-horizontaltable {
 
 .os-raise-text-bold {
   font-weight: 700;
+}
+
+.os-raise-img-fluid {
+  max-width: calc(100% - 1rem);
```
rnathuji, line 281: Just out of curiosity: This seems pretty subtle to me from a quick test. Is there a specific scenario where it makes a bigger difference, or just a good practice to create a bit of space?

src/styles/content.scss
```diff
@@ -275,4 +275,9 @@ table.os-raise-horizontaltable {
 
 .os-raise-text-bold {
   font-weight: 700;
+}
+
+.os-raise-img-fluid {
```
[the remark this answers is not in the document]
MReyna12: We could potentially use it for `<video>` as well, so that's one reason I kept it "open" (maybe something like `os.-raise-media-{somethingCleverHere}`. You also increase the specificity I believe by adding the `img` prefix. If we think it might get used outside of its use-case then I am open to narrowing it down to your suggestion.

src/styles/content.scss
MReyna12: I wouldn't say a big difference, but one particular example where it might be useful is shown below. I actually was going back and forth between just making it `100%`. Now that I've had time to sit on it + you noting it, I think that may be better and any spacing between content should be handled differently. I originally was also trying to mirror what was used inside of Moodle to try and keep some consistency.

Notice because of the `1 rem` the image doesn't touch the table before it wraps to the top.
https://github.com/openstax/k12-apps-raise/assets/37000585/dd95dbea-48cc-4396-b931-7034c882a6a9

src/styles/content.scss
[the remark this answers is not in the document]
rnathuji: That makes sense. I'm fine with leaving open (either with the current name or a different one based on your preference).

src/styles/content.scss
rnathuji: It's probably fine either way, but I'll defer to your judgment on what might be most consistent.

rnathuji (approved):

PR #133, opened 2023-11-28
Add styling for extra support boxes

MReyna12:


MReyna12:
You'll notice in the markdown that I did use `os-raise-extrasupport-title` in two places for the Support for English Language Learners example. Based on the Figma file Maddy provided, the same exact styling is applied for both pieces of text. I didn't update the name of the class as I thought it `title` was still applicable, but I am open to updating the class name if the reviewer thinks otherwise.

docs/styles.md
```diff
@@ -713,6 +715,76 @@ Div objects only
...
+        Use the first read to orient students to the situation by asking students to describe what the situation is about without using numbers (a teacher buys calculators and measuring tapes for her class).
+        Use the second read to identify quantities and relationships; ask students: “What can be counted or measured?”
+        Listen for, and amplify, the important quantities that vary in relation to each other in this situation: number of calculators purchased (for each order), number of measuring tapes purchased (for each order), and total cost of each order.
+        After the third read, reveal the questions that follow and invite students to brainstorm possible strategies to find the solution to the system.
+        This routine helps students interpret the language within a given situation needed to create an equation.
+      </p>
+      <p class="os-raise-text-italicize">Design Principle(s): Support sense-making</p>
+      <p class="os-raise-text-bold os-raise-extrasupport-title">Learn more about this routine</p>
```
rnathuji, line 757: Per [your comment](https://github.com/openstax/k12-apps-raise/pull/133#issuecomment-1830390570), the re-use of `.os-raise-extrasupport-title` here makes sense to me since UX wants the two to look the same. 

Howver, is the `.os-raise-text-bold` necessary if we're using `.os-raise-extrasupport-title` since the latter is setting `font-weight`? I think the latter will always win for rule specificity anyways :man_shrugging: .

MReyna12: Good catch! I'll remove the duplicate bold class.

rnathuji (approved):

PR #136, opened 2023-12-11
Update all px values to rem values

MReyna12:


src/styles/content.scss
```diff
@@ -16,10 +16,10 @@ ol.os-raise-noindent {
 }
 
 div.os-raise-graybox {
-  border: 1px solid $raise-graybox-border;
-  padding: 16px;
+  border: .063rem solid $raise-graybox-border;
```
rnathuji, line 19: Just curious: Is there a specific reason for rounding `1px` versus using `.0625`?

MReyna12: I used a px to rem converter to get `.063`. No other reason - I can change to `.0625`.

PR #137, opened 2023-12-15
Add student-reflection and utility styles

MReyna12:


docs/styles.md
```diff
@@ -26,16 +26,20 @@
 * Layout
   - Flex
     - Initialization of Flex Box
-      - [Required Flex Class](#side-by-side-os-raise-d-flex)
+      - [Required Flex Class](#side-by-side-os-raise-d-flex-nowrap)
```
rnathuji, line 29: I feel like we need to think through the doc changes a bit more in terms of structure. We've got two classes that can be used for flex box, but we now only document one (the other is referenced indirectly in other places like the table styles). This ToC indicates to me that this section is layout agnostic, but it takes me specifically to `Side by side (os-raise-d-flex-nowrap)`. There are probably cases when we want to use one vs the other, and we should help ensure that is made clear for authors.

On a related note: We have quite a bit of content already that is using `os-raise-d-flex` for "side-by-side". Unfortunately, those likely entail both intended scenarios (e.g. where we want wrapping and where don't), so there may not be a perfect solution in terms of being "backwards compatible". However, we may want to think things through a bit as far as the impact of effectively changing what we are using moving forward vs in the past.

docs/styles.md
```diff
@@ -26,16 +26,20 @@
...
-      - [Add Space Equally Around Side-by-Side Content](#side-by-side-os-raise-d-flex)
+      - [Center Side-by-Side Content Vertically](#side-by-side-os-raise-d-flex-nowrap)
+      - [Center Side-by-Side Content Horizontally](#side-by-side-os-raise-d-flex-nowrap)
+      - [Add Space Between Side-by-Side Content](#side-by-side-os-raise-d-flex-nowrap)
+      - [Add Space Equally Around Side-by-Side Content](#side-by-side-os-raise-d-flex-nowrap)
+    - Content Spacing
+      - [Add 16px of Space Between Content](#side-by-side-os-raise-d-flex-nowrap)
+      - [Add 32px of Space Between Content](#side-by-side-os-raise-d-flex-nowrap)
 * Components
   - [Gray Box](#gray-box-os-raise-graybox)
```
rnathuji, line 39: This isn't directly related to your PR, but maybe something we can fix real quick: The actual location of the `Gray Box` content does not reflect what is in the ToC. Can we move it so it is adjacent to `Motivcational Content`, etc.?

docs/styles.md
```diff
@@ -808,51 +843,67 @@ the side by side content.
...
 - `os-raise-justify-content-between` adds even spacing between content starting after the first element and ending before the last.
 - `os-raise-justify-content-evenly` adds even spacing between content including before the first and after the last element.
+- `os-raise-gap-1` adds 1 rem (16px) of space between content.
+- `os-raise-gap-2` adds 2 rem (32px) of space between content. 
 
 ```html
-<h3>os-raise-d-flex</h3>
-<div class="os-raise-d-flex">
+<h3>os-raise-d-flex-nowrap</h3>
+<div class="os-raise-d-flex-nowrap">
```
rnathuji, line 851: This has resulted in confusion in the past, and frankly confused me when I looked just now :sweat_smile: , but we should ensure that the HTML and examples are clearly tied together. In this case, the output of the HTML is in a completely different order from the image. Related to some of the other comments, maybe as part of improving the flex related documentation we should break things down very clearly with a 1:1 between image and HTML :man_shrugging: .

docs/styles.md
```diff
@@ -785,7 +789,34 @@ Support for Students with Disabilities
 
 ---
 
-## Side by side (os-raise-d-flex)
+## Student Reflection Box
+
+Add a styled container and content for student reflection.
```
rnathuji, line 794: Only because I think it will be useful for CMs, can we also put an example of the horizontal layout (e.g. the Building Character example)?

src/styles/content.scss
```diff
@@ -254,6 +254,18 @@ table.os-raise-horizontaltable {
   row-gap: 5rem;
 }
 
+.os-raise-d-flex-nowrap {
+  display: flex;
```
rnathuji, line 258: Minor: I know it's the default, but maybe it would be nice to explicitly say `flex-wrap: nowrap`?

rnathuji (commented):
@MReyna12 - The code changes here look good, but I'd like to think through some of the documentation a bit more to try and reduce confusion / improve clarity for CMs. We can discuss some of those comments tomorrow.

docs/styles.md
rnathuji: Just a quick follow-up (for discussion): Maybe we need to just carefully document uses of both `os-raise-dl-flex` and `os-raise-d-flex-nowrap` since some of this will be up to what behavior the CM wants. The table scenarios we have are a case in point, but I think in general the flex box section should be self-consistent + holistic in how it describes the various utility classes. :thinking:

MReyna12:
Overall, I am on board with updating the entirety of the layout section. I was originally going to restructure it and provide individual screenshots for each class, but since that seemed to maybe reach outside of the scope of this issue I just followed the pattern we currently have. I too found some of the pieces confusing - I should have raised a suggested update.

Happy to discuss further tomorrow.

docs/styles.md
MReyna12: My original thought was to document both `os-raise-d-flex` and `os-raise-d-flex-nowrap`. Once those two are established, then the next "step" would be to show how each of our utility classes work when paired with `os-raise-d-flex` and `os-raise-d-flex-nowrap`. If the behavior was the same, then we could state as much. If the behavior was different then we can show / explain said different behavior.

I also thought it might be particularly helpful to explicitly state that utility classes "A - F" will only work with one of the flex classes. It may also be useful to note that some utility classes work on the flex parent, while others only work on the flex "children" (aka the flex-items).

docs/styles.md
MReyna12: Yes, this can be done as well.

docs/styles.md
MReyna12: Good catch - my vote is to break things down more clearly with a 1:1 between the image and HTML. While the flex content can and should be connected (since they only work when paired together), individual images and HTML like we do for other content in the markdown would go a long way.

docs/styles.md
MReyna12: Yes.

src/styles/content.scss
MReyna12: Is the suggestion to explicitly state `flex-wrap: nowrap` because we have a flex class that does have wrap? The reason I ask is because I think there may be additional defaults that are a part of `display: flex` (like flex-direction) and want to make sure the intent isn't to identify all defaults.

docs/styles.md
MReyna12: Fixed.

src/styles/content.scss
rnathuji: Yeah, it was mostly just to aid a bit of readability. You make a good point, though, that maybe the specificity of just the one may make things muddier. I rescind this comment :joy: .

docs/styles.md
MReyna12: Added.

docs/styles.md
MReyna12: Documentation has been updated. There are now individual examples for each item as it relates to flexbox. If there is anything that doesn't seem to flow, let me know. Overall, I think this will hopefully provide more clarity on flexbox usage.

rnathuji: Thanks for the doc improvements, @MReyna12 ! They look great to me.

PR #138, opened 2023-12-18
Add accordion style

P-Gill97:


src/lib/styles.tsx
```diff
@@ -0,0 +1,49 @@
+export const stylify = async (element: HTMLElement): Promise<void> => {
```
rnathuji, line 1: This file can just be `.ts` vs `.tsx` since there is no React.

rnathuji, line 1: Minor: As of now at least, I don't think we need to make this `async` (it doesn't have any `await`s or anything). That would allow you to remove some of the `.catch()` when invoking.

src/styles/content.scss
```diff
@@ -335,4 +335,70 @@ table.os-raise-horizontaltable {
...
+  align-items: center;
+  background-color: white;
+  padding: 0.625rem;
+  border-bottom: 1px solid #D9D9D9;
+  color: $os-dark-green;
+  font-size: 1rem;
+  font-weight: bold;
+  border: none;
+  width: 100%;
+  cursor: pointer;
```
rnathuji, line 359: Randomly noticed this -- is it still necessary if we're using `<button>`?

src/styles/content.scss
```diff
@@ -335,4 +335,70 @@ table.os-raise-horizontaltable {
 .os-raise-extrasupport a {
   color: #0A3E5B;
   text-decoration: underline;
-}
\ No newline at end of file
+}
+
+.os-raise-accordion {
```
rnathuji, line 340: Wasn't sure where to put this comment, but I thought I'd flag something I noticed when navigating via keyboard: I see red bottom borders on FF (I would screenshot, but when I go to do it they go away :upside_down_face: ). On Chrome they seem to be black. Maybe those are defaults and there is some styling we need to specify?

src/styles/content.scss
rnathuji: Ah -- I just realized that we do actually need it. Things will work in Moodle otherwise because of [this](https://github.com/moodle/moodle/blob/513f3b02c76e0321b3c57ca8014d3100b4da8d94/theme/boost/style/moodle.css#L13466), but we need to specify it here so it works on the OER site.

src/styles/content.scss
rnathuji: :thinking: -- No, that's what I'm doing. Maybe I can screen share when we chat later today. Perhaps it is specific to my environment / build of FF.

rnathuji (approved):
This PR looks good to me to merge when you're ready, @P-Gill97 .

src/styles/content.scss
rnathuji: @P-Gill97 : Just as a follow up from our live convo, I confirmed that on my Mac I see the blue vs red when tabbing across the accordion sections.

rnathuji: :thinking: -- I see. Playing around a bit, I notice that in general the focus colors when navigating via keyboard vary across my systems (Linux vs Mac) on other items like math, interactives, etc. Given that, I think if we want to tweak these for consistency, it's probably something we need to look into holistically. I vote we leave this alone so we're consistently inconsistent for the time being :sweat_smile: .

rnathuji: @P-Gill97 - Yeah, good call as far as getting input from UX. I do think that if we want to create CSS to make things consistent, it's probably a separate issue we need to spin up so we can identify / address multiple items that I'm finding can vary.

PR #142, opened 2024-01-24
Remove compute engine

P-Gill97:


specs/blocks.spec.ts
```diff
@@ -308,7 +308,7 @@ test('attempts-exhausted response overrides answer-specific response', async ({
 test('Math field component rendered', async ({ page }) => {
   const htmlContent = `
   <div class="os-raise-ib-pset" data-schema-version="1.0" data-retry-limit="3">
-    <div class="os-raise-ib-pset-problem" data-problem-type="input" data-solution='3\\sqrt2+\\frac42' data-problem-comparator='math'>
+    <div class="os-raise-ib-pset-problem" data-problem-type="input" data-solution='3\\sqrt2+2' data-problem-comparator='math'>
```
rnathuji, line 311: Purely out of curiosity because it's not obvious to me :sweat_smile: : Why does the `\\frac42` fail now?

rnathuji (approved):
This PR looks good :+1: . I had one minor question for my own edification.

specs/blocks.spec.ts
rnathuji: Ah -- right :+1:

PR #143, opened 2024-01-30
Add styling for family support material links

MReyna12:


src/styles/content.scss
```diff
@@ -429,3 +429,25 @@ $arrow-svg: url("data:image/svg+xml,%3Csvg width='16' height='9' viewBox='0 0 16
...
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='52' viewBox='0 0 56 56'%3E%3Cpath fill='%230a5b50' d='M15.555 53.125h24.89c4.852 0 7.266-2.461 7.266-7.336V24.508c0-3.024-.328-4.336-2.203-6.258L32.57 5.102c-1.78-1.829-3.234-2.227-5.882-2.227H15.555c-4.828 0-7.266 2.484-7.266 7.36v35.554c0 4.898 2.438 7.336 7.266 7.336m.187-3.773c-2.414 0-3.68-1.29-3.68-3.633V10.305c0-2.32 1.266-3.657 3.704-3.657h10.406v13.618c0 2.953 1.5 4.406 4.406 4.406h13.36v21.047c0 2.343-1.243 3.633-3.68 3.633ZM31 21.132c-.914 0-1.29-.374-1.29-1.312V7.375l13.5 13.758Zm5.625 9.985h-17.79c-.843 0-1.452.633-1.452 1.43c0 .82.61 1.453 1.453 1.453h17.789a1.43 1.43 0 0 0 1.453-1.453c0-.797-.633-1.43-1.453-1.43m0 8.18h-17.79c-.843 0-1.452.656-1.452 1.476c0 .797.61 1.407 1.453 1.407h17.789c.82 0 1.453-.61 1.453-1.407c0-.82-.633-1.476-1.453-1.476'/%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: 1rem 50%;
+  background-size: 1.5rem;
+  min-height: 3.5rem;
+  padding-left: 3.5rem;
+  margin-bottom: 1rem;
+}
+
+.os-raise-familysupport .os-raise-familysupport-text {
```
rnathuji, line 448: Would it be too brittle to make the assumption that `.os-raise-familysupport` will have a single `<p>`? If so, maybe we could simplify slightly and remove `.os-raise-familysupport-text` to just make this `.os-raise-familysupport p` and then the schema doesn't need the additional class in `<p>`? :man_shrugging:

MReyna12: Realistically, there is nothing stopping a CM from adding a million `<p>`. I thought the `.os-raise-familysupport-text` class explicitly shows CMs how to achieve the "text" styling. However, I am open to removing `.os-raise-familysupport-text` and updating `styles.md` with language that states only a single `<p>` should be used to achieve the styling shown in the example. Thoughts?

rnathuji: True, but IDK how much of a concern that would be in practice given the limited use cases UX/PDM have in mind for this. I think not having the additional class makes this align better with other styles like `.os-raise-motivation`, `os-raise-graybox`, tables, etc. I know we had lots of classes in `.os-raise-extrasupport`, but that was really an outlier. For the most part, in our styles we've tried to take a "minimal markup"/single class approach wherever possible :man_shrugging: .

MReyna12: Updated the styling and documentation in this [commit](https://github.com/openstax/k12-apps-raise/pull/143/commits/ed0d67304fd7cbd6f3eef556bf675327d84628ee).

PR #150, opened 2024-03-20
Override pset block retry limit values

MReyna12:


MReyna12:
The test I put together confirms the problem level retry limits populate for each individual problem. If we need to do a deeper test to confirm overriding, then that might require testing question attempts up to the individual problem retry limit levels.

src/tests/ProblemSetBlock.test.tsx
```diff
@@ -496,3 +496,43 @@ test('ProblemSetBlock does not call the onProblemAttempt handler on default cont
...
+      <div class='os-raise-ib-pset-attempts-exhausted-response'>
+        <p>The answer is 42 </p>
+      </div>
+    </div>
+    <div class="os-raise-ib-pset-problem" data-retry-limit="3" data-problem-type="dropdown" data-solution="red" data-solution-options='["red", "blue", "green"]'>
+      <div class="os-raise-ib-pset-problem-content">
+        <p>Dropdown problem content</p>
+      </div>
+    </div>
+    <div class="os-raise-ib-pset-problem" data-retry-limit="4" data-problem-type="multiselect" data-solution='["red", "blue"]' data-solution-options='["red", "blue", "green"]'>
```
rnathuji, line 516: Just for completeness, can we include a problem that isn't overridden / inherits the retry limit from the parent?

src/components/DropdownProblem.tsx
```diff
@@ -31,14 +31,16 @@ export function buildClassName(response: string, correct: boolean, formDisabled:
```
rnathuji, line 31: Oh -- a general comment I almost forgot: Let's make sure to update docs for users in `schemas.md` in this PR too that describes how to use this optional override.

src/components/ProblemSetBlock.tsx
```diff
@@ -171,6 +173,7 @@ export const ProblemSetBlock = ({ waitForEvent, fireSuccessEvent, fireLearningOp
       contentId: prob.contentId,
       solution: prob.solution,
       retryLimit: prob.retryLimit,
```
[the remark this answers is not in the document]
MReyna12: It does work - I originally opted for being a little bit more explicit in lieu of DRYing things up. It may not be obvious if you only look at the problem types, but if you look for how they get their props I think it would not be hard to understand so long as you understand the logic behind `??`.

I've updated to match the pattern noted above.

src/tests/ProblemSetBlock.test.tsx
MReyna12: Added via latest commit.

src/components/DropdownProblem.tsx
MReyna12: Added via latest commit.

rnathuji (approved):

PR #152, opened 2024-04-08
Add style updates and update schema screenshots

MReyna12:


MReyna12:
Since there were some minor styling adjustments, I went ahead and added updated screenshots to styles.md.

PR #155, opened 2024-04-18
Add user message styles

P-Gill97:


rnathuji (commented):
Some quick comments on this:

1. Let's make sure to include doc changes similar to [here](https://github.com/openstax/k12-apps-raise/blob/main/docs/styles.md#family-support-materials-links) for these new styles so CM has clear examples + corresponding screenshots for reference.
2. I'm not sure that the CSS as defined here aligns with the usage pattern established by `.os-raise-familysupport`, but I think it'd be nice if the schema was the same for these two new boxes but just a different class name:

```html
<div class="os-raise-{something}">

</div>
```

3. Can we confirm / clarify whether the lightbulb box is going to have links? (the mockup in the issue doesn't seem to have one, but I'm not sure if it's a sometimes or always thing)

rnathuji:
I think that works / the class names seem reasonable to me. Tying back to the examples in the mockups they'd look something like:

```html
<div class="os-raise-usermessage-link">
  <p>Use this link to access the activity: <a href="">Desmos Activity: RAISE 3.5.1 Using Bivariate Data in Context</a></p>
</div>
<div class="os-raise-usermessage-lightbulb">
  <p>If you're having trouble seeing the x-intercept, try resizing the graphing window by pressing the '-' button in the upper right hand corner</p>
</div>
```

src/styles/content.scss
```diff
@@ -461,3 +461,45 @@ $arrow-svg: url("data:image/svg+xml,%3Csvg width='16' height='9' viewBox='0 0 16
...
+.os-raise-usermessage-link p,  .os-raise-usermessage-lightbulb p {
+  color: $raise-light-black;
+  font-weight: 700;
+  font-size: .875rem;
+  margin-bottom: 0;
+}
+
+.os-raise-usermessage-link a, .os-raise-usermessage-lightbulb a {
+  text-decoration: underline;
+}
```
rnathuji, line 505: We're introducing a bit of asymmetry here as it relates to styling links in these boxes:

image


Can we confirm with UX whether :point_up: is intentional, or if we should be adding `.os-raise-familysupport` here like:

```css
.os-raise-usermessage-link a,
.os-raise-usermessage-lightbulb a,
.os-raise-familysupport a {
  text-decoration: underline;
}
```

rnathuji (approved):

src/styles/content.scss
madisonschear: Yes they should be underlined. Thank you @rnathuji good catch!

PR #164, opened 2024-06-04
Migrate Jest to Vitest

P-Gill97:


package.json
```diff
@@ -50,5 +50,9 @@
     "vitest": "^1.6.0",
     "@vitest/coverage-v8": "^1.6.0",
     "whatwg-fetch": "^3.6.20"
+  },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "4.18.0"
```
rnathuji, line 55: @P-Gill97 - I saw this commit, and I think it is actually unnecessary. I suspect the build error was created by something that went a bit off when you merged `package-lock.json` in this PR (perhaps [this commit](https://github.com/openstax/k12-apps-raise/pull/164/commits/91f74c34cb70e5c48d0829b8c4e960a5410bfd68)). In lieu of this change, I would recreate `package-lock.json` using something like:

```bash
$ rm -rf node_modules package-lock.json
$ npm install
```

:point_up: should fix the build error you tried to address with this change to `package.json`.

PR #167, opened 2024-07-18
Update cta block interactivity and style

MReyna12:


src/styles/interactives.scss
```diff
@@ -137,6 +138,45 @@ div.os-justify-content-end {
...
+
+.os-submit-button-default-theme {
+  @extend %os-submit-button-base;
+
+  background-color: #757575;
+  border: 1px solid #757575;
+  box-shadow: 0 .125rem .25rem 0 hsla(0deg 0% 46% / 25%);
+}
+
+.os-submit-button-style-theme {
```
rnathuji, line 172: Would it be better to call this `.os-submit-button-green-theme`? The convention would then essentially look like `.os-submit-button-{themname}-theme` for if we add more in the future :man_shrugging: .

Same comment would apply to the `.os-cta-container-style-theme`.

docs/schemas.md
```diff
@@ -79,18 +81,38 @@ When the button is clicked, the component will:
...
+<div style="text-align: center;">
+    <img src="./static/cta_block_style.png" width="500">
+</div>
 
 Notes on schema:
 
 * The `data-wait-for-event` attribute is optional and where specified should correspond to a `data-fire-event` from another Interactive Block on the same page (e.g. as part of a Content Template)
 * The `data-fire-event` attribute is optional
 * The `data-button-text` is optional but allows the content developer to customize the text in the CTA button. The default value is `Next`.
+* The `data-style-theme` attribute is optional and when specified will provide the styling shown above.
```
rnathuji, line 115: It might be nice to list the "supported themes" (currently it'd just be `green` of course)

src/components/CTABlock.tsx
```diff
@@ -31,19 +32,37 @@ export const CTABlock = ({ content, prompt, buttonText, fireEvent, waitForEvent
...
   const maybeButton = (): JSX.Element => {
     if (!clicked) {
-      return (<>
-      <button onClick={clickHandler} type="button" className="btn btn-outline-primary mb-3">{buttonText}</button></>)
+      return (
+        <>
+          <button
+            onClick={clickHandler}
+            type="button"
+            className={ctaStyleTheme === "green" ? "os-submit-button-style-theme" : "os-submit-button-default-theme"}>
```
rnathuji, line 40: I think the trailing `>` here can be removed.

src/components/CTABlock.tsx
```diff
@@ -31,19 +32,37 @@ export const CTABlock = ({ content, prompt, buttonText, fireEvent, waitForEvent
 
   const maybeButton = (): JSX.Element => {
     if (!clicked) {
-      return (<>
-      <button onClick={clickHandler} type="button" className="btn btn-outline-primary mb-3">{buttonText}</button></>)
+      return (
```
rnathuji, line 35: Could we DRY this up with:

```javascript
    return (
      <>
        <button
          onClick={clickHandler}
          type="button"
          className={ctaStyleTheme === "green" ? "os-submit-button-style-theme" : "os-submit-button-default-theme"}
          disabled={clicked}>
          {buttonText}
        </button>
      </>
    )
  }
```

specs/content-templates.spec.ts
```diff
@@ -31,7 +31,7 @@ test('segmented content template works', async ({ page }) => {
     <p>Want to see more?</p>
 </div>
 </div>
-<div class="os-raise-ib-cta" data-button-text="Yes!" data-fire-event="event2" data-wait-for-event="event1">
+<div class="os-raise-ib-cta" data-button-text="Yes!!" data-fire-event="event2" data-wait-for-event="event1">
```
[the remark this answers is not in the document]
MReyna12: Previously, we were removing the button, so when the test would look for the `Yes!` button the second time there was only one button showing `Yes!`. Now that we are instead disabling the button, there were two `Yes!` buttons and the test was failing because there were two `Yes!` buttons that could be clicked.

src/lib/blocks.tsx
```diff
@@ -163,6 +164,7 @@ export const parseCTABlock = (element: HTMLElement): JSX.Element | null => {
     buttonText={buttonText}
     fireEvent={fireEvent}
     waitForEvent={waitForEvent}
+    ctaStyleTheme={maybeCTAStyleTheme}
```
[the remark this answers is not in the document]
MReyna12: No we wouldn't. Updated.

src/styles/interactives.scss
MReyna12: Funnily enough that was the first name I gave it, but wasn't sure if we wanted to keep using "style-theme".

Updated both.

src/components/CTABlock.tsx
MReyna12: I think we need to keep the trailing `>` because it closes `<button`.

specs/content-templates.spec.ts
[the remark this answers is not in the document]
rnathuji: Ah, right. Thanks!

src/components/CTABlock.tsx
rnathuji: Ah, right. My :eyes:  missed that.

docs/schemas.md
```diff
@@ -79,18 +81,39 @@ When the button is clicked, the component will:
...
+* Supported values for `data-style-theme`: `green`.
```
rnathuji, line 116: Minor: Would indenting this help with clarity?

docs/styles.md
```diff
@@ -1267,6 +1269,50 @@ Any element that contains text.
...
+</div>
+
+**Availability**
+
+Any element that contains text and is part of CTA block content.
+
+**Usage**
+
+```html
+<p class="os-raise-cta-title-style-theme">This text is green, bold, and 24px.</p>
```
rnathuji, line 1311: I think you may have missed updating the HTML here with the name change.

MReyna12: I did - must have misspelled when I did my search across the files.

rnathuji (approved):

PR #169, opened 2024-07-23
Update dependency versions

rnathuji:


MReyna12:
Point 1: Totally fine with turning them off at this stage of the game.

Point 2: I also agree that these deprecation warnings will have to be addressed by Bootstrap.

P-Gill97 (approved):
Tested on local instance. Everything seems to work!

PR #170, opened 2024-07-23
Update general input box styles

P-Gill97:


docs/schemas.md
```diff
@@ -162,7 +162,13 @@ If students are entering something into a text block, be sure to follow these co
 The output if the `data-wait-for-event="eventnameY"` attribute is removed and the user submits an answer.
```
MReyna12, line 162: Based on this text, I would expect the image that follows to depict the status of the UI when a user submits an answer. You've opted for showing active, submitting without providing a response, and "correct" submission. Do you think we should add additional text to clarify the two additional states of the `UserInputBlock`?

src/styles/bootstrap.scss
```diff
@@ -171,15 +172,19 @@
     /* stylelint-disable-next-line scss/at-extend-no-missing-placeholder */
     @extend .form-control;
 
-    border: .125rem solid $os-default-answer-border-color;
+    border: .125rem solid $os-default-gray;
     border-radius: .3125rem;
-    box-shadow: 0 .125rem .125rem $os-default-box-shadow;
+    box-shadow: 0 .125rem .125rem $os-default-gray;
```
MReyna12, line 177: The border and box-shadow values should mirror the values I've added via my PR. If I merge after you, then it should be no issue. This is what I have:

`
border: .063rem solid #757575;
border-radius: .3125rem;
box-shadow: 0 .125rem .25rem 0 hsla(0deg 0% 46% / 25%);
`

src/styles/bootstrap.scss
```diff
@@ -171,15 +172,19 @@
...
   }
 
   input.os-form-control {
     padding: .75rem;
   }
 
+  textarea.os-form-control {
+    padding: 1rem 1.5rem;
+  }
+
```
MReyna12, line 187: We were previously letting bootstrap determine the padding for the `textarea`, right? If that's the case then this makes sense. I didn't see anywhere else where we were setting the padding for said element.

src/styles/bootstrap.scss
P-Gill97: Ahh that makes sense! For some reason, the Figma file wasn't showing me the box shadow and border numbers! I went back and apparently, I had to click on the input box a bunch of times. Did figma give you the hsla value?

P-Gill97: Wait I think I see it now, Did you select the HSL from the dropdown when looking at the properties?

src/styles/bootstrap.scss
P-Gill97: The acceptance criteria says "Change input form padding 24 left/right, 16 top/bottom".  Did I interpret that incorrectly?

src/styles/bootstrap.scss
MReyna12: Getting the box shadow down was weird because the values provided by Figma were not translating to CSS. I guess in Figma they select the color they want and then can add a second color option that lightens or darkens the primary color. So I got the default color and looked at how much Maddy wanted the color adjusted. I think it was 25 percent, so with those values I came up with the next color for the box-shadow. 

Yeah, some of the values aren't available unless you click through the nested elements. Even then they are typically in pixels so have to adjust things to rem anyways.

src/styles/bootstrap.scss
P-Gill97: Yeah I think before It was defaulting to some value provided by bootstrap.

src/styles/bootstrap.scss
P-Gill97: Ok that makes sense!!

docs/schemas.md
P-Gill97: I'll go back to having the submitted state only,  to be more consistent with what we had and make the documentation less verbose.

MReyna12 (approved):

PR #178, opened 2024-09-16
Update default content version

MReyna12:


rnathuji (approved):
Looks good to me. I confirmed the content version sha by loading [https://k12.openstax.org/contents/raise/8170a112/smoketest.json](https://k12.openstax.org/contents/raise/8170a112/smoketest.json).

